PR #81, opened 2013-07-25
Adds SemVer Comparisons

onefifth:
Allows for SemVers to be compared against each other and adds support for SemVers with >, <, <=, >= and = modifiers to define open ended version ranges.
This could eventually be leveraged when defining and validating package dependencies versions in haxelib projects.

### DIFF
--- a/haxelib.json
+++ b/haxelib.json
@@ -4,7 +4,7 @@
 	"license": "GPL",
 	"tags": ["haxelib", "core"],
 	"description": "The haxelib client",
-	"version": "3.1.0-rc.2",
+	"version": "3.1.0-rc.3",
 	"releasenote": "Fixed issues with `haxelib git` and `haxelib dev`.  Added `classPath` to haxelib JSON format. Small fixes.",
 	"contributors": ["back2dos", "ncannasse", "jason"]
 }
--- a/haxelib.json
+++ b/haxelib.json
@@ -4,7 +4,7 @@
 	"license": "GPL",
 	"tags": ["haxelib", "core"],
 	"description": "The haxelib client",
-	"version": "3.0.0-rc.7",
-	"releasenote": "Added 'haxelib convertxml' and improved 'haxelib git' support.",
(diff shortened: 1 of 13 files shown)


onefifth:
Commit 0cfc177 is related to this.

nadako:
I have no idea what to do with this one, and feel like closing it for the time being.

ibilon:
While important, just checking if a semver is in a "semver set" is not enough,
we need to be able to specify a set in haxelib.json,
and something a lot smarter than right now to use (#290).

Otherwise something like this, we want to install something depending on libA and libB:
libA depends on libC `(>= 1.0.0 && < 1.1.0) || (>= 1.2.0 && < 1.3.0)`
libB depends on libC `>= 1.0.0 && < 1.2.0)`
then libA will try to install the latest valid libC (by projecting on how that's currently done) say 1.2.5,
libB will do the same and end up with libC 1.1.7,
and our lib cannot be used (should have installed something like 1.0.1).

Alright this example was crated to fail and is very specific, it can still happen.

I think we need more design on this, and I do think it's important, before implementation.

onefifth:
This was originally written as purely an implementation giving haxelib the ability to compare semvers with the intent of extending it into the kind of dependency solving @ibilon describes above. When this PR was written, haxelib was using string comparison and did not have a way to even compare lib versions at all.

Lots has changed, there is now a basic implementation of semantic versioning in place and this PR is turning into much higher level discussion about how we want to solve dependencies rather than just defining and comparing.

If some minor work should be done to improve this solution, or at the least make this merge-able again, I would be happy to contribute if some decisions were made on exactly what needs to be accomplished.

nadako:
I think we shouldn't proceed with merging this, but continue design discussions (not now tho, but after 3.3 release). Too bad Github doesn't have an option to convert a PR to normal issue.

MrClogsworthYT:
hi am i late

Simn:
At this point I'll close this old PR.

PR #94, opened 2013-09-27
Fix Invalid Operation (+) error when env vars are missing.

m22spencer:
HOMEDRIVE/HOMEPATH are not always guaranteed to be present, if they are not, haxelib throws "Invalid Operation (+)".
http://stackoverflow.com/questions/18909393/trying-to-install-openfl-haxelib-returns-invalid-operation

As a more permanent solution to this, should we store haxelib config data in AppData? This is the correct location for it. If not, could we switch to USERPROFILE?

### DIFF
--- a/src/tools/haxelib/Main.hx
+++ b/src/tools/haxelib/Main.hx
@@ -675,9 +675,15 @@ class Main {
 		var haxepath = Sys.getEnv("HAXEPATH");
 		if( haxepath != null )
 			haxepath = Path.addTrailingSlash( haxepath );
-		var config_file;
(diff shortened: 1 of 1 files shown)


nadako:
I added some checks to avoid "Invalid operation (+)" error in b56fa4e3922282fd459cad018eab13491a029fe5 and added a new issue about using APPDATA - #259. Closing this PR.

ashes999:
I ran into this today. To anyone else who finds this paper-trail: it's fixed in Haxe 3.3.0.

PR #128, opened 2014-04-04
Location-independent haxelib runscript.

ioga:
One can create a symlink to that script and run it from any place
without bothering about proper classpath etc.

### DIFF
--- a/haxelib_script.sh
+++ b/haxelib_script.sh
@@ -1,2 +1,4 @@
 #!/bin/sh
-exec haxe --run tools.haxelib.Main "$@"
+HAXELIB_PATH=$(dirname $(readlink -f $0))
+CLASSPATH=$HAXELIB_PATH/src
+exec haxe -cp $CLASSPATH --run tools.haxelib.Main "$@"


nadako:
@andyli is this still relevant? I'm on Windows, so I can't check this out, but I think haxe's `make install` generates this script nowadays.

ioga:
Yes, the issue is still there. In https://github.com/HaxeFoundation/haxe/blob/development/Makefile#L114 one can currently see:
`echo "exec haxe -cp $(INSTALL_LIB_DIR)/extra/haxelib_src/src --run haxelib.client.Main \"\$$@\"" >> $(INSTALL_BIN_DIR)/haxelib`

i.e. the installation directory is hardcoded into haxelib script, and therefore you cannot move it around, and staged installs mentioned in https://github.com/HaxeFoundation/haxe/issues/2850 still won't work.

This particular changeset (as well as some part of haxe#2850) is probably outdated though since it's almost 2 years old.

andyli:
The "haxelib script" is problematic in many ways.. So I'm very tempted to remove it and use the good old nekotools booted haxelib instead. But I have to figure out this implication to the selfupdate thing... which is another source of headache to me.

nadako:
See #172 as @ibilon proposed a way to handle selfupdate where we can just keep bundled haxelib as a `nekoboot`ed executable.

nadako:
I'll close this as we seem to be moving away from shell script to nekobooted executable (see #172 and #284)

PR #135, opened 2014-04-30
Download dependencies from GIT (and haxelib) with a new entry : "requirements"

Peekmo:
Hello,

This PR introduce a new way (more flexible) to declare dependencies in haxelib.json file
To not break current libraries, I added a new key "requirements", which is an array that contains informations about these dependencies.

You can have a look to a test JSON file here : https://github.com/Peekmo/Peekmo-haxe-tools/blob/master/haxelib.json

Why ? Because in a company, you don't always want to share your libraries on haxelib, and for moment, that's impossible to download dependencies from a GIT project, and more, that's impossible to declare dependencies to git project inside a haxelib project. And that's my need, to be able to share my tools without adding them to haxelib (because they are not really interesting outside my projects).

My idea is (perhaps) not the best, but with this kind of declaration, we could support subversion for example (just for example XD), or something else, who know the future ? 

The "version" for a git project => git tag

Thank you for your reading.

### DIFF
--- a/src/tools/haxelib/Data.hx
+++ b/src/tools/haxelib/Data.hx
@@ -56,10 +56,10 @@ typedef GitPackageInfos = {
 }
 
 typedef RequirementsInfos = {
-	var name 			   : String; // Library's name
-	var type 			   : String; // Library's type (git | haxelib)
-	@:optional var version : String;
-	@:optional var package : GitPackageInfos; // Package's infos
+	var name 			      : String; // Library's name
+	var type 			      : String; // Library's type (git | haxelib)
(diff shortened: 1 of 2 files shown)


fzzr-:
This is really cool. Imho need to support hg and create tests for this feature.

nadako:
How about we support this similar to npm? See https://docs.npmjs.com/files/package.json#dependencies. I personally find it quite concise and convenient.

nadako:
I'll close this PR as we need to discuss format first and this branch has conflicts anyway. Let's move discussion to #238.

Issue #146, opened 2014-06-20
haxelib improvements


ibilon:
Hi,

a friend and I started working on a package manager for haxe because we feel that haxelib is missing a fair amount of functionality.
But after discussing with @hughsando it might be better to try and get them added here instead.

So here's a list of what we think is missing, some might already be work in progress in that case ignore them ;) some would be easy to fit in the existing code, while others may not.

1. Real dependency management: do not allow to remove a lib if this will break all the lib that depend on it
2. Haxe/neko as package: way easier updating, and people would actually know they have updates available
3. Multiple repository: for things like local enterprise repo or for people to easily provide betas or nightly build
4. Adding a field like "dependencies" but that don't add the libs when compiling
5. A field containing the name of the command alias that would be stored in a single bin folder: for things like lime or nme without needing custom code or sudo on linux
6. Post-install script
7. Integrity check
8. Having the possibility to only download the ndll you need: downloads are slow outside of europe
9. Fuzzy dependency: required version could be 1.0.0-keyword.1 , 1.0.0 , 1.0.* or 1.*
10. Better support for beta version: most likely done with an optional "beta" or "testing" repository (canal), or just proposing version with "-alpha/beta/rc" if an option is checked
11. the path command returns too many information, would benefit from a command that actually return the library path
12. run script gets the path they were called from as argument, which is a problem when you want to also use the run.n manually, while developing it for instance. Using an environment var would be less intrusive
13. the ability to install the packages globally or locally for the user, like npm/pip/... do

Multiple repository and integrity check would require a different architecture, similar to the one used by linux package manager:
having a metadata file, very small, containing only a timestamp and the hash of the next file: fileslist which would contain all the zip available, their hash and size.
These two files would be cached on the user computer and used to process dependency offline and then present the user with the list of what need to be downloaded, could also allow for parallel downloading.
Like I said very similar to a standard linux package manager.

I know some people are really missing some of these features (cc @jgranick),
what do you think?

back2dos:
Sorry, but I'm going to have to close this. It would really help if you guys could make individual issues of the parts that really matter. Thanks.

PR #147, opened 2014-06-21
Add where command

ibilon:
Hi,

This add the _where_ command that print the path to the libraries given in arguments.
Proposition 11 from https://github.com/HaxeFoundation/haxelib/issues/146,
would be useful when you write tools that need the path of a library without having to parse the output of the `haxelib path` command.

For the help message I copied the one from _path_,
but I think that the message from _path_ should be the one to change since it doesn't print libraries paths but their (and their dependencies) classpath/ndll path/-D.
Not sure what to replace the message with though.


Simn:
I'm closing this because the PR cannot be merged anymore and because I don't see the usefulness of this. If anything this should be solved via `path`.

PR #171, opened 2015-03-05
Support haxelib upgrade without "haxelib selfupdate"

jgranick:
There are other ways to solve this problem, but the following is a solution that allows a single haxelib script to be compiled to an executable and stored in the Haxe directory, while also using the same executable elsewhere, installed using the "haxelib_client" lib.

Currently, upgrading haxelib is a hassle. On Windows (for example), you use:

    haxelib selfupgrade
    haxe update.hxml

...and this may very well fail. It does on my system.

Instead of replacing "haxelib.exe" (which is brittle), I believe that the standard haxelib executable should work as a shim to run the upgraded version of haxelib. This is trickier, since the haxelib repository may not be setup, yet, but this system tries to run the "haxelib_client" version if it exists with a haxelib.n binary, and otherwise uses itself to process the commands.

This could allow haxelib to be updated using "haxelib upgrade" like other libraries. Similarly, you could "haxelib dev haxelib_client path/to/haxelib_client/clone" and work on haxelib directly.

I'm not sure if we should try using "haxe --run" instead of using Neko to call the haxelib.n binary, I wasn't immediately sure how "haxe --run" worked with arguments so I opted for this solution instead. A "haxe --run" solution could be nice because it would work without recompiling the haxelib client code (not sure if that's good or bad, though, if you're working on the file)

### DIFF
--- a/src/tools/haxelib/Main.hx
+++ b/src/tools/haxelib/Main.hx
@@ -151,7 +151,30 @@ class Main {
 
 	function new() {
 		args = Sys.args();
-
+		
+		if (args.length == 0 || (args[0] != "-redirect" && args[0] != "setup")) {
+			try {
+				var rep = getRepository ();
+				var list = new List ();
+				checkRec ("haxelib_client", null, list);
+				var d = list.pop ();
+				if (d != null) {
+					var pdir = Data.safe(d.project)+"/"+Data.safe(d.version)+"/";
+					var dir = rep + pdir;
+					try {
+						dir = getDev(rep+Data.safe(d.project));
+						dir = Path.addTrailingSlash(dir);
+						pdir = dir;
+					} catch ( e : Dynamic ) { }
+					var n = pdir + "/bin/haxelib.n";
+					if (FileSystem.exists (n)) {
+						var code = Sys.command ("neko", [ n, "-redirect" ].concat (args));
+						Sys.exit (code);
+					}
+				}
+			} catch (e:Dynamic) {}
+		}
+		
 		commands = new List();
 		addCommand("install", install, "install a given library, or all libraries from a hxml file", Basic);
 		addCommand("upgrade", upgrade, "upgrade all installed libraries", Basic);
@@ -264,6 +287,7 @@ class Main {
 			if( a == null )
 				break;
 			switch( a ) {
+			case "-redirect":
 			case "-debug":
 				debug = true;
 			case "-notimeout":


ncannasse:
That seems like a bad idea for several reasons, in particular it requires neko to be installed while on Linux we are running haxelib through haxe internal interpreter. We could generalize this behavior by having haxelib.exe be a shortcut for  "haxe -lib haxelib_client -main Main --run ....", but it might be easier to fix the issue you're having if you can report it.

jgranick:
Sure, I meant to open this as a source of discussion. Instead of calling Neko, it could call `haxe --run`

The point of this is that once someone has Haxe and haxelib installed, they can "haxelib upgrade" and receive a new version of haxelib -- no "selfupgrade" required. They can also "haxelib dev haxelib_client path/to/haxelib_client/clone" and immediately begin to work on an improved version of haxelib.

It would be possible for "haxelib" to be a shortcut for "haxelib_client", but wouldn't this fail if haxelib has not been setup yet on the system?

jgranick:
Another option would be to have two scripts, one for haxelib_client, one for haxelib -- the latter being a shim that knows how to run "setup" and otherwise calls haxelib_client, this is how the Lime tools work (with run vs. the real tools) to enable in-place upgrades and rebuilds of the tools

waneck:
Selfupdate is a little difficult to test. But IIRC it's not working when
running through haxe interpreter

jgranick:
In general, I think "selfupdate" is a bad idea, the way it's implemented.

Getting "haxe" and "haxelib" may require administration rights, and once in place, I think it would be best to leave it. I think that "haxelib" should be a pass-through to a real library installed deeper (like "haxelib_client")

This allows "haxelib upgrade" to upgrade itself, it allows "dev" or "git" versions of haxelib, I think it makes the tool much stronger.

ncannasse:
@jgranick could you please report the error you're getting ? that will be much much MUCH more helpful than starting a debate about whereas it's a good implementation or not :)

ncannasse:
Regarding administrative rights : sadly I think the same apply to the .hx than to the .exe, unless we choose to put haxelibs into the local use Documents directory, which is another big change a bit outside of the scope here

jgranick:
When a program on your system is using haxelib (FlashDevelop, and invisible process running somewhere, so on) you will not be allowed to replace haxelib.exe, it will be an access violation

jgranick:
It is also difficult to test a new version of haxelib, in order to make improvements

ncannasse:
@jgranick haxelib should never block, it always print its result then exit, I fail to see how FlashDevelop would be bale to lock it

jgranick:
`haxelib run`

PR #177, opened 2015-03-12
haxelib version and help

andyli:
Added two simple commands:
 * `haxelib version` prints the currently using haxelib version.
 * `haxelib help` prints haxelib usage.

I'm not sure if we should add alias in the form of flags, like `-h/--h/-help/--help` and `-version/--version`.


back2dos:
To be honest, I have just skimmed this, but it looks straight forward enough ;)

Issue #186, opened 2015-03-17
Using haxelib as a haxelib


hughsando:
Creating an extra process and parsing data seems like overkill when we have a haxelib sitting there in a haxelib and it makes me sad.
I suggest the following:
Move 99% of the code out of "Main" and put in in Haxelib, with main, just:
```
  var haxelib = new Haxelib(Sys.args());
  haxelib.process( );
 
 And Haxelib.hx

 public function new( args )
 public function getRepository( ?setup : Bool ) 
 public function function checkRec( ...)
```
Even making new and checkRec and args public would make this work as a library.

back2dos:
This is planned for obvious reasons. 

I've discussed it with @jasononeil and we really want to decouple the CLI and the underlying client logic. It's not really reasonable to have to pass in your arguments as an array, or have it just access your stdin if there's not enough of them.

If you are happy with the current interface, you can just use `@:privateAccess` for the time being ;)

hughsando:
Yes, this is fair enough.
I guess that as a library, a "read only" api might be simple.  The only one I would really use is "where is this library[:version]".  A List/Array of {name,version,isCurrent,directory,dependency[]} would be icing on the cake. Not too far off checkRec.

nadako:
Let's merge this with #186

ibilon:
@nadako wrong link ;)

nadako:
Oops, I probably meant #217 :)

PR #197, opened 2015-04-03
[feature] Mercurial (hg) support

fzzr-:
Implemented support for Mercurial (hg) : clone, diff, pull, update.
All behaviors for haxelib-cli look like similar for `haxelib git`-command.
In the future it will help to support many any version control systems.

For test:
`$ haxelib hg signal https://bitbucket.org/fzzr/hx.signal develop "" 0.9.2` and
`$ haxelib update signal`.

hg4haxelib opt

### DIFF
--- a/src/tools/haxelib/Main.hx
+++ b/src/tools/haxelib/Main.hx
@@ -793,9 +793,10 @@ class Main {
 				case Haxelib:
 					doInstall(d.name, d.version, false);
 				case Git:
-					doGit(d.name, d.url, d.branch, d.subDir, d.version);
-				//case Hg & Any Vcs:
-					//doVcs(d.name, d.url, d.branch, d.subDir, d.version);
+					doVcs(VcsID.Git, d.name, d.url, d.branch, d.subDir, d.version);
+				//TODO: add mercurial-dependency type to schema.json (https://github.com/HaxeFoundation/haxelib/blob/master/schema.json#L38)
+				case Mercurial:
+					doVcs(VcsID.Hg, d.name, d.url, d.branch, d.subDir, d.version);
 			}
 		}
 	}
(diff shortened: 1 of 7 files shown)


src/tools/haxelib/Main.hx
```diff
@@ -252,7 +254,7 @@ class Main {
 			if (cats[i] == null) cats[i] = [c];
 			else cats[i].push(c);
 		}
-		
+
```
back2dos: Any reason for this? We eventually want to make haxelib usable as a library (see #186), so making this global seems a step in the wrong direction.

src/tools/haxelib/Main.hx
```diff
@@ -951,39 +953,35 @@ class Main {
...
-			if (!gitExists())
-				throw GIT_UNAVAILABLE;
+
+		//TODO: get content from `.current` and use vcs only if there "dev".
+
+		var vcs:Vcs = Vcs.getVcsForDevLib(rep + "/" + p);
+		if(vcs != null)
+		{
+			if(!vcs.exists)
+				throw VCS_UNAVAILABLE;
```
back2dos: Given that this is a constant, the displayed error will not be helpful. I think it would be better to throw the `vcs` (allowing for a `catch (e:Vcs) {}`) or have an explicit error for that.

src/tools/haxelib/Main.hx
```diff
@@ -1287,18 +1263,29 @@ class Main {
 	function doGit(libName: String, gitPath: String, ?branch : String, ?subDir: String, ?version:String) {
 		var rep = getRepository();
 		var proj = rep + Data.safe(libName);
-		var libPath = proj + "/git";
 
+		// find existing repo
```
back2dos: There seems to be quite a lot of duplication between this and `doHg`. It also seems inconsistent to make `update` a responsibility of the `Vcs`, while `install` isn't. Could you try to structure this better?

back2dos:
This is a good start. Personally, I think it would be great if you could manage to abstract the VCS specific stuff away into the respective implementations, as far as possible (without jumping through hoops that is).

More importantly: since this project is in such a lousy state, what would really help a lot is if you could include tests for this (ideally also for the case where the vcs in not available).

src/tools/haxelib/Main.hx
fzzr-: `static public function ask( question )`
Ok, I'll move it to something like Cli or CliUtil.
`print`& `command` too.

src/tools/haxelib/Main.hx
fzzr-: This const never displayed as error if [Main.hx#L954-L956](https://github.com/fzzr-/haxelib/blob/feature/hg/src/tools/haxelib/Main.hx#L954-L956).
Okay. Do you want me to throw the instance of Vcs? O_o
I think i have to move `VCS_UNAVAILABLE` into Vcs and throw this const from it.
Or will create ENum VcsError maybe.

src/tools/haxelib/Main.hx
fzzr-: Yes. Thx. I see it and will do little refactoring :)

fzzr-:
@back2dos Thx for review. I'll update this branch and all will be beauty :100:%.

src/tools/haxelib/Main.hx
fzzr-: Wow. My mistake. I'll fix it.

fzzr-:
I'll write tests soon.

fzzr-:
I was break Git :(. I'll fix it.

fzzr-:
@back2dos ready for merge.

fzzr-:
@back2dos : Ready for merge. Please review.

back2dos:
It looks really quite good, but I currently cannot accept it yet, because of #204

As much as I like macros, I do think that what `Vcs.staticRegistration()` is doing could be quite easily achieved without such heavy artillery. I haven't thought it through, but I do wonder whether having a `[VcsID.Git => new Git(), VcsID.Hg => new Hg()]` isn't simple enough.

fzzr-:
@back2dos OK?

back2dos:
Indeed :)
I'll merge this tonight!

back2dos:
Merged by hand, as it conflicted with the symlink pull request.

back2dos:
Oh, and thanks ;)

fzzr-:
Thank you.

fzzr-:
@back2dos Oh, please approve #205 - this is important.

Issue #202, opened 2015-04-11
[Bug] remove <version> - removing symlink as dir


fzzr-:
```log
$ haxelib remove nme-dev 1.3.2
Error rmdir /Users/ak/Developer/Haxe/libs/nme-dev/1,3,2/project/unpack/SDL2_mixer-2.0.0/Xcode/Frameworks/FLAC.framework/Headers
```

Here `Headers@` & `Resources@` is a symlinks:
```shell
$ cd /Users/ak/Developer/Haxe/libs/nme-dev/1,3,2/project/unpack/SDL2_mixer-2.0.0/Xcode/Frameworks/FLAC.framework/
FLAC.framework ak$ ls -GFAC
Headers@          LICENSE.FLAC.txt  Resources@        Versions/
```

fzzr-:
@back2dos : You can assign this issue to me :)

fzzr-:
Fixed in https://github.com/HaxeFoundation/haxelib/pull/203.

back2dos:
Fixed by #203

PR #203, opened 2015-04-12
fixed issue #202 - removing symlinks

fzzr-:


fzzr-:
@back2dos : How should I add tests to appveyor.yml?

back2dos:
The appveyor calls haxe tests.hxml, which ultimately ends up here: https://github.com/HaxeFoundation/haxelib/blob/master/test/HaxelibTests.hx#L49

fzzr-:
Hmmm... What you think about https://github.com/fzzr-/haxelib/commit/a10c235789a93ebebe85de5352d3ea0ab41d417b#diff-354f30a63fb0907d4ad57269548329e3R9 ? :(

back2dos:
Uhm, this really isn't my strong suit, so I have to ask: why not just add it in the `HaxelibTests`? Am I missing something?

fzzr-:
My mistake. I'll do it very soon.

waneck:
I hate how our sys API is lacking a more sane way to detect this.

>

test/tests/TestRemoveSymlinks.hx
```diff
@@ -0,0 +1,41 @@
...
+		super();
+	}
+
+
+	//--------------- initialize ----------------//
+
+	override public function setup():Void
+	{
+		Sys.command("neko", ["./bin/haxelib.n", "setup", Sys.getCwd() + "haxelib-repo"]);
+		Sys.command("neko", ["./bin/haxelib.n", "install", "nme-dev", "1.3.2", "--always"]);
```
andyli: Please do not depends on 3rd party libs in our test. They could take long time to download. 
Create only a minimal lib in `/test/libraries` and use that.

back2dos: Well, I think using something smaller than nme-dev should not be that much of an issue, e.g. mcli, or whetever else you find under 100KB.

fzzr-:
__Ready for merge.__
@waneck : I agree and I have some ideas for std..

PR #206, opened 2015-04-23
path command can look in multiple locations for haxelibs

yanhick:
Hi,
this is a hack that we use internally to be able to search multiple paths for haxelibs. We only need it for the "path" command right now.
Here is how we use it:
```HAXELIB_PATH=/haxelib/path:/other/haxelib/path haxelib path mylib```

Obviously this is a very limited implementation only serving our use case but I'm wondering:
- Did I miss an existing similar feature or is it in the roadmap ?
- else is there interest for a full implementation of this feature (meaning for all commands) ?

### DIFF
--- a/src/tools/haxelib/Main.hx
+++ b/src/tools/haxelib/Main.hx
@@ -936,17 +936,27 @@ class Main {
 	}
 
 	function getRepository( ?setup : Bool ) {
+		return getRepositories(setup).pop();
+	}
+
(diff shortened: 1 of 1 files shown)


Simn:
I'm not opposed to this in principle, but addressing it for path specifically is not a good approach. This requires a larger review of how repositories are handled in the client.

I'm closing the PR because it doesn't merge. If anyone has strong feelings about this feature please open a normal issue.

PR #247, opened 2015-09-23
Propagate --always and --never answer to dependencies

### LINKED ISSUE #246: --always and --never options are not propagated to dependency install commands
It appears that if you use `--always` and `--never` flags to control the `[y|n|a]` prompts, these options are not propagated to dependency installs in certain situations.

Briefly looking at the code, it seems like it could be related to the `Vcs`-related class(es) creating new instances of `Cli`, which have `defaultAnswer` set to null by default, or the initialization order of the main `Cli` instance with the `Cli` instances owned by `Vcs` instances.

```sh
> haxelib --never git thx.promise git@github.com:fponticelli/thx.promise master src
Installing thx.promise from git@github.com:fponticelli/thx.promise
Cloning into '/Users/awhite/temp/haxelib-test/.haxelib/thx,promise/git'...
remote: Counting objects: 576, done.
remote: Total 576 (delta 0), reused 0 (delta 0), pack-reused 576
Receiving objects: 100% (576/576), 99.83 KiB | 0 bytes/s, done.
Resolving deltas: 100% (206/206), done.
Checking connectivity... done.
Library thx.promise set to use Git.
  Branch/Tag/Rev: master
  Path: /Users/awhite/temp/haxelib-test/.haxelib/thx,promise/git/src
Installing dependency thx.core
You already have thx.core version 0.34.0 installed
Set thx.core to version 0.34.0 [y/n/a] ? n   #### should not have prompted here
```

andywhite37:
- Possibly resolves #246 
- The issue was that when the Vcs class creates new Cli() instances, it
  clears the defaultAnswer value that was parsed from the command line
arguments at startup.  Not setting defaultAnswer to null in the Cli
constructor allows you to preserve the defaultAnswer (stored in the
static defaultAnswer_global var that backs defaultAnswer).
- I'm not sure if this change will have other unwanted side effects

### DIFF
--- a/src/tools/haxelib/Main.hx
+++ b/src/tools/haxelib/Main.hx
@@ -144,10 +144,8 @@ class Cli
 	public var cwd(get_cwd, set_cwd):String;
 	private static var cwd_cache:String = null;
 
-
-	public function new()
-		defaultAnswer = null;
-
+	public function new() {
+ 	}
 
 	function get_cwd():String
 	{


Simn:
Wow this is some confusing stuff. An instance variable that has a setter which assigns to a static variable. What could possibly go wrong?

Anyway, in that context the change makes sense to me.

PR #251, opened 2015-10-25
Ignore ".dev" if version specified (close #249)

### LINKED ISSUE #249: compiler -lib: ignore .dev if library version specified
I think, if a library version specified then dev library version must not be used:
```cmd
haxe -lib mylib:1.0.0 ...
```
So, file `.dev` must be ignored by compiler in this case.

yar3333:


yar3333:
Do I need to fix more in some other places? Is compiler run haxelib.n to know about paths or use internal code?

Simn:
This makes a lot of sense to me. I can't tell if your changes address everything though...

PR #258, opened 2015-12-09
Fix #250

### LINKED ISSUE #250: [MacOS X El Capitan] haxelib setup default location is unwritable
Running `haxelib setup` suggests to install on default location `usr/lib/haxelib` which is, as of Mac OS X El Capitan, unwritable due to the new "Rootless" feature.

It should probably point to install on the writable location `usr/local/lib/haxelib`.

jakwings:


nadako:
Thanks, I've merged it locally.

PR #268, opened 2015-12-20
Use safe project name for the zip download path (website)

### LINKED ISSUE #221: libraries with a dot in the name cannot be downloaded
Also the README.md doesn't show up for them probably because of the same issue.

ibilon:
Fix #221

In the cli https://github.com/HaxeFoundation/haxelib/blob/master/src/haxelib/Data.hx#L194-L196


nadako:
We could just use that `Data.fileName` method since it's a shared code, not CLI-related I think.

ibilon:
True, #269

PR #269, opened 2015-12-20
Use safe project name for the zip download path (website)

ibilon:
2nd try :)


nadako:
Thanks!

ibilon:
Also I'm not sure the `#deploy` special case is still required,
since in #221 the error message was `[file_open,/data/haxelib/www/../www/files/3.0/perf.js-1,0,2.zip]` @jasononeil ?

PR #270, opened 2015-12-20
Fix FileSystem.fullPath error when directory doesn't exist

ibilon:
If you try to do `haxelib dev` on a directory that doesn't exist `FileSystem.fullPath` will throw a `std@file_full_path` error before the `exists` check is run.


nadako:
I moved the `fullPath` call out of the try block (cee23ba574bb3d94ad213df296fec2c405f01087) to prevent misleading "couldn't write to devfile" errors in case when fullPath fails.

Issue #282, opened 2016-03-09
Available version check


ibilon:
I'm pretty sure (from experience) that most people don't know about self-update and end up using the bundled haxelib until the next haxe version.

We should really indicate to the user when a new version of haxelib is available.
```
A new version (X.Y.Z) of haxelib is available.
Do "haxelib selfupdate" to get the latest version.

/normal output/
```

The test would be quick, getting something like `http://lib.haxe.org/latestVersion/haxelib` which would contain only the semver (so very light).
It would be at least on the commands that already connect to the internet, and we should probably cache the result too.

That way we can really send bug fixes and improvements.

I think it's important to ship that feature with haxe 3.3.

nadako:
Good idea! Though I think it would be easier to implement a method in SiteApi instead of a http request.

ibilon:
Forgot about that but yes.

ncannasse:
Sounds like a good idea, but not on "path" please :) (used for every compilation)

nadako:
Yeah, as discussed and implemented in #283, it's done only for `net` commands

nadako:
Done in b779136d257be75aa8b839d9f92c44bd9b56ab1b (BTW @ibilon: you can use `closes #<issuenum>` in commit messages to automatically close issues when it gets merged)

ibilon:
I know, just completely forgot about this issue even though I opened it :)

Thanks for closing it.

Issue #172, opened 2015-03-05
Running and upgrading haxelib 


ncannasse:
After a discussion with @jgranick, I agree that we need to make sure that users can easily upgrade haxelib so we can introduce new usages without waiting for a new haxe release.

Maybe we should always use the haxe macro system on all OS (including Windows), and make sure that we can update the client by saving it into a directory we are sure the user has write access. I'ld surely like to hear what people think about good locations for that ?

jgranick:
I propose that C:\HaxeToolkit\haxe\haxelib.exe should become a small shim.

If the user has not setup haxelib, it should be able to find a default install of haxelib_client, included with Haxe. This could be located under C:\HaxeToolkit\haxe\std\tools\haxelib_client for example.

If the user has setup haxelib, it should look for a "haxelib_client" package in the haxelib local repository, for example, C:\HaxeToolkit\haxe\lib\haxelib_client\3,2,1

In the Lime and OpenFL tools, I use a simple script like this:

    var args = Sys.args ();
    var code = Sys.command ("neko", [ "tools.n" ].concat (args));
    Sys.exit (code);

I believe @ncannasse said that not being dependent on Neko would be preferable, so `haxe --run` against the current haxelib_client should also work.

*Benefits*

You would be able to update the haxelib_client automatically using `haxelib upgrade` like other libraries.

You could `haxelib dev haxelib_client path/to/dev/directory` to work on improvements to haxelib

You would not need to use `sudo` access on Mac or Linux to replace the "haxelib", it would be a shim that does not need to be replaced for an update.

I would be happy to help implement this

back2dos:
I will see to it that haxelib becomes a .bat on windows, that just launches a haxelib.n, which we can replace while running.

The `haxelib dev haxelib_client` approach sounds nice, but that wouldn't work since in order to run `haxelib` we'd need to call `haxelib path haxelib_client`, so that's kind of a chicken and egg problem.

jgranick:
So currently, haxelib.exe is the haxelib.n + Neko boot.

I was thinking, the shim could use Neko boot, but could use the Neko "loadModule" code in order to find the haxelib_client.n file. This would not require Neko on the system, nor would it require the Haxe executable (helping to avoid a chicken-and-egg problem)

[...]

PR #283, opened 2016-03-09
Notification of available haxelib version

ibilon:
Implementation of #282 

I added the check to all command that connects to the internet already (with the exception of install which don't always do).


ibilon:
Ok so maybe not a safe assumption that "haxelib_client" always exists :)

Simn:
On an unrelated random note, either you or Mark have to change your GitHub avatar or I'll start talking to the wrong person soon...

src/haxelib/SemVer.hx
```diff
@@ -97,6 +97,12 @@ abstract SemVer(String) to String {
 	inline function get_valid()
 		return isValid(this);
 
+	@:op(a > b) static function greaterThan(a:SemVer, b:SemVer) {
```
nadako: I think this is wrong, since it doesn't check `preview`/`previewNum` fields. There's `Semver.compare(a,b)` that will return 1 if a > b.

src/haxelib/client/Main.hx
```diff
@@ -154,27 +154,27 @@ class Main {
 		args = Sys.args();
 
 		commands = new List();
-		addCommand("install", install, "install a given library, or all libraries from a hxml file", Basic);
-		addCommand("update", update, "update a single library (if given) or all installed libraries", Basic);
+		addCommand("install", checkUpdateThen.bind(install), "install a given library, or all libraries from a hxml file", Basic);
```
nadako: I think this should be done differently. We have `net` property for commands, so the check should be simply added after `loadProxy` here: https://github.com/HaxeFoundation/haxelib/blob/development/src/haxelib/client/Main.hx#L366

ibilon:
Thanks for the feedback @nadako 
I'll have to look at the source code, seems I'm missing quite some knowledge about haxelib.

nadako:
Np. I wonder though if it's better to add `SiteApi.getLatestVersion` that just returns a version string instead of using `infos` that returns the whole project information. Concerning both traffic and server db load.

ibilon:
I thought about that but I was unsure on the effect of modifying SiteApi, even when just adding something.

nadako:
It shouldn't break anything, I think, but like any API, it should be carefully designed. To me, it feels a bit weird to have `getLatestVersion` for what's already possible to do with the `infos` method, but OTOH it has a specific goal: to return latest library version in the most efficient way, so maybe it's okay.

I'll just ping @Simn, @ncannasse and @back2dos and hear what they think. :)

ibilon:
I checked just to be sure, and yes adding a function doesn't break an old client.

nadako:
Right, update also uses `infos().getLatest()` which can be replaced by the new shiny method. I'd say let's do it!

ibilon:
Did a try on it, had to go sql for that since I couldn't figure out how to order on several column otherwise.

src/haxelib/client/Main.hx
```diff
@@ -1037,17 +1048,17 @@ class Main {
 		}
 		else
 		{
-			var inf = try site.infos(p) catch( e : Dynamic ) { Sys.println(e); return; };
-			p = inf.name;
```
nadako: This one was needed, actually, see f358bfada2812970f94986defe6f2108d232dfe2, gotta look into how to get that name in proper case without `infos` call.

src/haxelib/server/Repo.hx
```diff
@@ -80,6 +80,15 @@ class Repo implements SiteApi {
 		};
 	}
 
+	public function getLatestVersion( project : String ) : SemVer {
+		var p = Project.manager.select($name == project);
+		if( p == null )
+			throw "No such Project : "+project;
+
+		var v = Version.manager.unsafeObjects('SELECT * FROM Version WHERE project == ${p.id} ORDER BY major DESC, minor DESC, patch DESC, previewNum DESC LIMIT 1', false);
```
nadako: I think this is not correct, since it doesn't account `preview` field that should be also considered for sorting. Not sure if it can be done within a simple SQL query.

src/haxelib/server/Repo.hx
```diff
@@ -80,6 +80,15 @@ class Repo implements SiteApi {
 		};
 	}
 
+	public function getLatestVersion( project : String ) : SemVer {
+		var p = Project.manager.select($name == project);
+		if( p == null )
+			throw "No such Project : "+project;
+
+		var vl = Version.manager.search($project == p.id, { orderBy: [-major, -minor, -patch, -previewNum], limit: 1});
```
nadako: Same comment as for the previous version:
> I think this is not correct, since it doesn't account preview field that should be also considered for sorting. Not sure if it can be done within a simple SQL query.

ibilon: Oh right, for some reason I thought those where the same field.
Will have to check some sql, no idea on how enums work there.

src/haxelib/client/Main.hx
ibilon: Oh so there was a reason, thought it was weird to do that if the value was identical.
It a problem though :/ and adding a `realName` function to SiteApi just for that seems suboptimal, would have the merit to be explicit at least. Wouldn't happen to be any other places that do that?

ibilon:
```
var latest = try site.getLatestVersion(p) catch( e : Dynamic ) { Sys.println(e); return; };
p = p.toLowerCase();
p = FileSystem.readDirectory(pdir).filter(function (dir) return dir.toLowerCase() == p)[0];
```
that would be too ugly I guess?

If only the folder was made in lower case to begin with :/

nadako: You probably meant `rep` instead of `pdir`, but yeah, that might work! Please factor that code into a separate function tho :)

src/haxelib/server/Repo.hx
nadako: Gotta check SPOD macros, maybe enum indices are used, then you can sort on them easily.

ibilon:
Updated getLatestVersion, had to go back to sql.

They are indeed stored as int,
so you can easily sort them, or you should but they are nullable and null is below all int,
but 3.2.0 > 3.2.0-rc.

Sql has a function `isnull(var, val)` which replace var with val if it's null,
can use the same values as SemVer.compare => 100 for preview and -1 for previewNumber (so not needed).

But sqlite named it `ifnull` otherwise that'd be too easy :)

nadako:
Fails OSX now: https://travis-ci.org/HaxeFoundation/haxelib/jobs/114881349#L3845

ibilon:
Fixed it, apparently isnull is something, works with ifnull on both mysql and sqlite (that's even better actually).

andyli:
I agree it would be useful, but please don't merge it until someone has high confidence that selfupdate actually works for all configurations.
If haxelib keeps telling ppl to do selfupdate when it actually doesn't work it would be funny... ; )

nadako:
Right, let's deal with selfupdate in #172

ibilon:
With #284 should the message be changed to `Do "haxelib update haxelib_client to get the latest version`?

(update installs the latest version even if the lib isn't installed)

Since that's basically all that `selfupdate` would do.

Personally I'm not fan of redundant commands like upgrade/update, install/local/

nadako:
One more issue to think about: currently we duplicate version number in `haxelib.json` and in `Main.VERSION` which is prone to mistakes. E.g. you fix stuff, change `haxelib.json`, upload a new version, but forgot to update the `VERSION` field, and you'll be getting a message about updating to the current version, which is just plain retarded).

I see two immediate solutions for this:
 * Load the `VERSION` value from `haxelib.json` using macro.

or vise-versa...

 * Build `haxelib.json` from a template in our `Package.hx` script, reading version from `Main.VERSION`.

I think second one is the most error-proof, since we eliminate the case where we first build the `run.n` with the old `haxelib.json` version and then change `haxelib.json` :)

src/haxelib/SemVer.hx
```diff
@@ -97,6 +97,10 @@ abstract SemVer(String) to String {
 	inline function get_valid()
 		return isValid(this);
 
+	@:op(a > b) static inline function greaterThan(a:SemVer, b:SemVer) {
```
nadako: It feels a bit weird to only have `>` operator and not the other ones. For a single case, maybe it's simplier to just use `compare` method directly instead of `>`, even though it doesn't look great... I'm not sure though.

src/haxelib/client/Main.hx
```diff
@@ -188,6 +188,14 @@ class Main {
 	}
 
 
+	function checkUpdate() {
+		var latest:SemVer = try site.getLatestVersion("haxelib_client") catch (_:Dynamic) null;
+
+		if (latest != null && latest > VERSION) {
+			print('A new version (${latest}) of haxelib is available.\nDo "haxelib selfupdate" to get the latest version.\n');
```
nadako: Don't forget to reword this to `haxelib update haxelib` as we want to deprecate `selfupdate`

src/haxelib/client/Main.hx
```diff
@@ -188,6 +188,14 @@ class Main {
 	}
 
 
+	function checkUpdate() {
+		var latest:SemVer = try site.getLatestVersion("haxelib_client") catch (_:Dynamic) null;
```
nadako: The library is going to be `haxelib` instead of `haxelib_client` as were discussed.

nadako:
I added a couple comments here. Could you please rebase and squash this?

src/haxelib/client/Main.hx
```diff
@@ -173,8 +173,8 @@ class Main {
 		addCommand("local", local, "install the specified package locally", Development, false);
 		addCommand("dev", dev, "set the development directory for a given library", Development, false);
 		//TODO: generate command about VCS by Vcs.getAll()
-		addCommand("git", function()doVcs(VcsID.Git), "use Git repository as library", Development);
-		addCommand("hg", function()doVcs(VcsID.Hg), "use Mercurial (hg) repository as library", Development);
+		addCommand("git", function() doVcs(VcsID.Git), "use Git repository as library", Development);
+		addCommand("hg", function() doVcs(VcsID.Hg), "use Mercurial (hg) repository as library", Development);
```
nadako: unrelated whitespace change, but oh well...

src/haxelib/SiteApi.hx
```diff
@@ -26,6 +26,7 @@ import haxelib.Data;
 interface SiteApi {
 	public function search( word : String ) : List<{ id : Int, name : String }>;
 	public function infos( project : String ) : ProjectInfos;
+	public function getLatestVersion( project : String) : SemVer;
```
nadako: looks like a space is missing before the `)` here :)

src/haxelib/client/Main.hx
```diff
@@ -1037,19 +1049,32 @@ class Main {
...
+				setCurrent(p, latest, true);
 		}
 	}
+
+	function projectNameToDir( project:String, state ) {
+		var p = project.toLowerCase();
+
+		var l = FileSystem.readDirectory(state.rep).filter(function (dir) return dir.toLowerCase() == p);
+
+		switch (l.length) {
```
nadako: it's not important or anything, but note that this could be written as
```haxe
switch (l) {
    case []: throw ...;
    case [dir]: return Data.unsafe(dir);
    case _: throw ...;
}
```
the generated code will be pretty much the same, but I personally find it more readable :)

src/haxelib/client/Main.hx
ibilon: Modified it and then returned it back, thought the spaces where from before, I'll return it to normal in the squash.

src/haxelib/SemVer.hx
ibilon: Would it be less weird if I added the other five?

nadako: It seems reasonable to have comparison operators for SemVer, though I usually follow "refactor is used more than twice" rule :) I don't have a strong opinion here...

ibilon:
Successfully squashed without changing pull request this time :)

Although I see that I brought run.n

nadako:
Nice!

> Although I see that I brought run.n

That's how it should be now, since we want haxelib to be easily installable directly from git too :) Good thing it doesn't change between recompiles if source hasn't changed.

nadako:
Merged by hand since it had conflicts with my earlier changes, sorry!

ibilon:
Np ;)

nadako:
> (update installs the latest version even if the lib isn't installed)

Turns out you broke this with `case []: throw "No such project " + project + " installed"` in the `projectNameToDir` function. I changed that to simply `return project`.

Issue #235, opened 2015-07-23
-notimeout has no effect after message : "Sending data..."


wighawag:
without -notimeout:
```
haxelib submit lib.zip
Http connection timeout. Try running haxelib -notimeout <command> to disable timeout
```

with -notimeour
```
haxelib -notimeout submit lib.zip
Password :
Sending data....
Http connection timeout. Try running haxelib -notimeout <command> to disable timeout
```

ibilon:
The timeout parameter is ignored for the next step (processing) https://github.com/HaxeFoundation/haxelib/blob/master/src/tools/haxelib/Main.hx#L654 but normally not for this one.

Maybe the error message is too broad https://github.com/HaxeFoundation/haxelib/blob/master/src/tools/haxelib/Main.hx#L482 and the error is not timeout related? (just guessing here)

juef17:
This seems to be an issue with `install` as well:

```
C:\Users\juef>haxelib install openfl
Downloading openfl-3,5,3.zip...
Download complete : 0 bytes in 10.5s (0KB/s)
Http connection timeout. Try running haxelib -notimeout <command> to disable timeout

C:\Users\juef>haxelib -notimeout install openfl
Downloading openfl-3,5,3.zip...
Download complete : 0 bytes in 10.5s (0KB/s)
Http connection timeout. Try running haxelib -notimeout <command> to disable timeout
```

nadako:
Looks like `-notimeout` only has effect on remoting calls, not on general http request, for which it should be done using `Http.cnxTimeout` field.

nadako:
This should work now.

PR #286, opened 2016-03-10
Don't ignore -notimeout in submit

ibilon:
See #235 

submitting shouldn't remove the effect of `-notimeout`


nadako:
Thanks!

nadako:
See https://github.com/HaxeFoundation/haxelib/issues/235#issuecomment-194928792

PR #287, opened 2016-03-10
Don't install version-less dependencies that have a dev set

### LINKED ISSUE #174: `haxelib install` should not install dependencies if they exist as a "dev" or "git" library already
When you run `haxelib install`, dependencies are installed, even if they already exist.

For example, if you have OpenFL and Lime set up as `haxelib dev` directories, installing HaxeFlixel will force the download of both libraries before finishing the install.

ibilon:
Fix #174


nadako:
Thanks!

Issue #272, opened 2015-12-22
haxelib run + cwd behaviour


nadako:
The current behaviour is quite weird: We append cwd to the end of argument list and then change cwd to the library installation directory. What's the reason it's implemented like that? cc @ncannasse 

I think it would be more correct not to change cwd and set the lib directory to the `HAXELIB_RUN` env var (currently it's always set to `1`).

Changing this will break current haxelibs, so if we want to do it, we should think of some way to stay compatible.

nadako:
Assigning to @ncannasse for comments :)

ncannasse:
I agree with the change. Not sure how to make sure to handle backward compatibility.

nadako:
One possibility is to add some "haxelib protocol version" field to haxelib.json and use the new mechanism if it's set to proper value (and warn if author submits a library using old mechanism). But I'm not sure it's worth adding that field just for this change.

PR #291, opened 2016-03-11
Make self update work on top of run mechanism

ibilon:
https://github.com/HaxeFoundation/haxelib/pull/284#issuecomment-195353626

### DIFF
--- a/client.hxml
+++ b/client.hxml
@@ -1,3 +1,3 @@
 -cp src
--neko bin/haxelib.n
+-neko run.n
 -main haxelib.client.Main
--- a/src/haxelib/client/Main.hx
+++ b/src/haxelib/client/Main.hx
@@ -157,6 +157,10 @@ class Main {
 	{
 		args = Sys.args();
 
+		if (Sys.getEnv("HAXELIB_RUN") == "1") {
+			Sys.setCwd(args.pop());
+		}
+
 		commands = new List();
 		addCommand("install", install, "install a given library, or all libraries from a hxml file", Basic);
 		addCommand("update", update, "update a single library (if given) or all installed libraries", Basic);
@@ -348,18 +352,16 @@ class Main {
 			}
 		}
 
-		if (!settings.safe && FileSystem.exists(getRepository() + "haxelib_client")) {
+		if ((!settings.safe && Sys.getEnv("HAXELIB_RUN") != "1") && FileSystem.exists(getRepository() + "haxelib_client")) {
 			var old_argcur = argcur;
 			argcur = 0; // send all arguments
-			args.unshift("--safe");
 
 			try {
-				doRun("haxelib_client", null, false);
+				doRun("haxelib_client", null);
 			} catch (e:String) {
 				if (e.startsWith("Library haxelib_client version ") && e.endsWith(" does not have a run script")) {
 					// old version of haxelib_client without run script, ignore
 					argcur = old_argcur;
-					args.shift();
 				} else {
 					throw e;
 				}
@@ -1374,7 +1376,7 @@ class Main {
 		doRun(temp[0], temp[1]);
 	}
 
-	function doRun( project:String, version:String, changeDir=true ) {
+	function doRun( project:String, version:String ) {
 		var rep = getRepository();
 		var pdir = rep + Data.safe(project);
 		if( !FileSystem.exists(pdir) )
@@ -1391,10 +1393,8 @@ class Main {
 			catch (e:Dynamic)
 				throw 'Error parsing haxelib.json for $project@$version: $e';
 
-		if (changeDir) {
-			args.push(Sys.getCwd());
-			Sys.setCwd(vdir);
-		}
+		args.push(Sys.getCwd());
+		Sys.setCwd(vdir);
 
 		var callArgs =
 			if (infos.main == null) {
--- a/src/haxelib/client/Main.hx
+++ b/src/haxelib/client/Main.hx
@@ -157,7 +157,7 @@ class Main {
 	{
 		args = Sys.args();
 
-		if (Sys.getEnv("HAXELIB_RUN") == "1") {
+		if (Sys.getEnv("HAXELIB_RUN") != null) {
 			Sys.setCwd(args.pop());
 		}
 
@@ -352,7 +352,7 @@ class Main {
 			}
 		}
 
-		if ((!settings.safe && Sys.getEnv("HAXELIB_RUN") != "1") && FileSystem.exists(getRepository() + "haxelib_client")) {
+		if ((!settings.safe && Sys.getEnv("HAXELIB_RUN") == null) && FileSystem.exists(getRepository() + "haxelib_client")) {
 			var old_argcur = argcur;
 			argcur = 0; // send all arguments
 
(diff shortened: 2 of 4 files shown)


nadako:
Ooh, you should've just force-pushed into previous PR branch, that PR has some interesting discussion, but oh well.

andyli:
Thanks a lot for the work!
I've just reviewed the design, but haven't looked into the code yet. It sounds pretty good to me.

I can see that there are still issues. I describe them below, followings with a proposal to mitigate.

 1. This PR's implementation depends on how `haxelib run` works. What if we want to change that in the future? e.g. It has been discussed we may replace "run.n" with a "Run.hx" that is run by the macro interpreter, such that we can remove the neko dependency.
 2. The existing haxelibs may not be able to perform selfupdate. We should document the issue, but I guess we cannot assume everyone will read...

My proposed solution: Implement a way to prevent people with certain old haxelib version downloading new versions of "haxelib_client". We can let the haxelib client add an additional GET param that states it version when downloading libraries. The server will return an error if the requesting client has an version incompatible to the new version. For existing haxelib clients, they will not send out this GET param, thus the server will know that it must be an old client.

What do you think? If you agree and don't have another design, you may merge this PR and we can start to work from here.

nadako:
> This PR's implementation depends on how haxelib run works. What if we want to change that in the future? e.g. It has been discussed we may replace "run.n" with a "Run.hx" that is run by the macro interpreter, such that we can remove the neko dependency.

`haxelib run` already supports interp version, if you specify `main` field in the `haxelib.json` file, that should be enough. I didn't really review the `changeDir` stuff in the PR, it may have some issues if we're going to change the behaviour regarding cwd in `run` (see #272).

> The existing haxelibs may not be able to perform selfupdate. We should document the issue, but I guess we cannot assume everyone will read...

Actually, on windows I think they will be able, since it nekoboots a new executable. The problem is with unixes that was discussed in #284

> My proposed solution: Implement a way to prevent people with certain old haxelib version downloading new versions of "haxelib_client". We can let the haxelib client add an additional GET param that states it version when downloading libraries. The server will return an error if the requesting client has an version incompatible to the new version. For existing haxelib clients, they will not send out this GET param, thus the server will know that it must be an old client.

I wonder if that can/should be implemented by using our "api version" mechanism. This requires more thought as we clearly don't want to introduce a lot of code for such migration.

ibilon:
Point one: Indeed we are dependent of how run works, but on the other hand we use it here because we know it's working really well.
If we change how it works then haxelib is just like all affected libraries. 

Point two: no they can, and it'd work as long as they keep the first "new" haxelib the upgrade to, unless it's broken already for them (bundled nekobooted version from an haxe installer on linux) but nothing we can do here.

Interesting solution, so filtering the result of `SiteApi.infos` based on the transmitted version (or lack thereof) of haxelib and if an error is thrown it'll get printed, maybe too much just for that, and not sure how we'd actually do it.

The `changeDir` is true by default and will do the same thing as before, it's set to false when running the haxelib_client so that it's behaving just like if you launched it from the directory where you launched haxelib.

nadako:
Okay, I looked at the `changeDir` logic. It feels a bit weird to use different run mechanism here.

Maybe it would be better to conform the current run protocol and check for `HAXELIB_RUN` env variable, pop the cwd from the last argument and do `Sys.setCwd` again. That way we can run with both `haxelib run haxelib_client` and just `haxelib`. We also don't need to add `--safe` arg when calling `haxelib_client` from haxelib, just check for EITHER `--safe` or `HAXELIB_RUN` env var (if we do that, we can rename `--safe` to something more intuitive, but i'm not sure what exactly atm).

Personally, I'm not a fan of this cwd dance, but at least it would be consistent. And we have another issue about reworking that (#272).

nadako:
As for forbidding updates, another option would be to just rename the library, so it won't be updated for older clients, but newer ones will check for the new name. I think I like this approach, since it's not like people are expecting selfupdate to work anyways :)

ibilon:
For the changeDir, that would indeed allow to let run as is, I'll change it.

I like the renaming approach, and it would fully work, how about just `haxelib`?

nadako:
> I like the renaming approach, and it would fully work, how about just haxelib?

Ok with me

nadako:
@andyli do you think this gets it all covered?

nadako:
Hmm, I wonder how big extra process calling overhead will be for `haxelib path` runs. They are quite performance-critical because they are called by haxe compiler for each lib.

ibilon:
We could "fix" it by having only one call to `path` per compilation, for all libs mentioned, and all their deps.

Would be even better actually, well for the next haxe at least.

nadako:
Agreed, we should look into that.

ibilon:
We could also load the `neko.n`, that's not the run command at all anymore, but that'd avoid a new process.
No idea if that's faster, or safe, though.

nadako:
You mean `run.n`? Yeah I also thought about that, I also have no idea if it's any better :)

ibilon:
Yeah :)

Well the file is load from disk and parsed in both case,
so that's the cost of a new process versus the one of the code to load a n module in another.

Maybe not worth the risk now that I'm thinking more about it.

ibilon:
Oh I was worried that people without haxe 3.3+ wouldn't get the fix and ends up with a high cost for path on compilation, but nobody should have haxelib 3.3+ without haxe 3.3+ (since it doesn't selfupdate to it) or haxe 3.3+ without haxelib 3.3+ (that would be weird).

So it'll be ok, could even change the default behavior of path I guess.

nadako:
Yeah, now would be the right time to implement support for `haxelib path` multiple arguments.

nadako:
I'll look into it now.

nadako:
Oh wait, it's already there and actually works just like that, sorry for the noise!

nadako:
And it's cached for completion, so I think this overhead doesn't really matter.

nadako:
I'm itching to merge this \o/

ibilon:
The lib name isn't changed yet, should other people vote on it? Not like it's a big deal.

nadako:
I don't think anyone besides Andy cares as long as it works :)

andyli:
Yeah, I think the plan is all good now.

I like renaming "haxelib_client". I understand the existing haxelibs *may* selfupdate correctly, but they may not as well, so I think it is good to force everyone to get the new haxelib alongside with haxe 3.3, and selfupdate will work reliably from now on.

I still think that it would be useful to send additional params when the clients request to install new libs. It could gives us statistics about the number of currently active haxe/haxelib versions etc. But it could be implemented later.

So, please go ahead to change the `doRun("haxelib_client")` to `doRun("haxelib")` (and other related parts), and then merge as you wish : )

nadako:
Okay, then I'll merge this and then change library name and cleanup leftovers.

After that I think there's some work to do regarding installers, linux/homebrew packages and whatnot. Will you handle that @andyli?

andyli:
No problem, I will handle those. It is merely changing the haxe Makefile to use a nekobooted haxelib instead of the scripted version.

nadako:
Merged manually in b4799e889bcacccb852f7b56a06bd49c71032b42

andyli:
Oh, one more consideration:
When haxe 3.4/4 is released and bundle a even newer haxelib. The user's "haxelib haxelib" may still exist with an older version. So the newer haxelib may call the older haxelib... Should there be a sam-ver check to ensure the newer one is always prefered (except `--safe`)?

nadako:
Hmm, maybe we should check if installed version greater than current!

nadako:
I made a meta-issue with current stuff related to this: #293

PR #294, opened 2016-03-12
Assign a different exit code to each error

ibilon:
It's not good practice to return just 0 or 1.

Documentation:
```
# Return codes (doesn't apply to haxelib run)
0. No error
1. Missing directory argument for -cwd
2. -cwd point to an unvailable directory
3. Both --always and --never are set
4. Unknown command
5. Cannot resolve host
6. Http connection timeout
7. Current working directory is unavailable
8. Unknown error
9. Trying to remove haxelib without --safe
10. Missing vcs executable
11. Couldn't clone vcs repository
12. Couldn't checkout vcs branch, tag or path
13. Couldn't checkout vcs tag
14. No haxelib.xml file found in current directory
```
no idea where to put it


nadako:
Sorry, I just committed something that conflicts with this...

ibilon:
That happens, but `run.n` is a merge conflict too (always since it's binary), more of a problem.

nadako:
Anyway, while I agree that it's a good idea in general, I'm not sure we should do it now, since this is basically a public API that people might depend on and I'd love to have some space for refactorings for now.

Also, I wonder if there's some widely-accepted convention of common exit codes that we could follow.

ibilon:
Agreed, look's like bash has a couple of reserved exit code we might want to not throw.

We'll have to have a serious discussion about the refactoring once  haxe 3.3 is out,
right now it's almost a fairytail :D

Issue #293, opened 2016-03-12
Issues with the new update mechanism


nadako:
I'll make a list of issues found with the new haxelib update mechanism, so we don't forget about them:

- [x] `haxelib remove haxelib` fails (at least on Windows) because the directory with `run.n` file is busy. One option would be to always use bundled haxelib if we're removing `haxelib`, but that sounds pretty hacky.
- [x] When user upgrades Haxe with a newer bundled haxelib, his installed haxelib may be of lower version than the bundled one. In general, this is a not big issue, since haxelib will show upgrade notification (#282), but this may be a problem if some newer Haxe version requires a newer haxelib version. We could check if installed haxelib version is greater than the bundled one.
- [x] I still think that we gotta check if there's a noticeable performance regression with `haxelib path`, since now we do some extra work (extra fs queries, reading `haxelib.json`, invoking `neko run.n`). However I wouldn't be surprised if it'll perform on par or even faster on Unixes because `haxe --interp` mode we used earlier is slower than Neko.
- [x] Should haxelib haxelib be searched in a project-local repository (created with newrepo)? I'd say not.
- [x] Check if it works correctly when haxelib repo path is not configured (no `/etc/.haxelib` or `~/.haxelib` file, no `HAXELIB_PATH` env var)
- [x] Document the usage of `--safe` option if haxelib installation is messed up

cc @ibilon

andyli:
For, `haxelib remove haxelib` i think it is reasonable to require the use of `--safe` option.

For haxelib haxelib & project-local repository, that's an interesting question... I think "do not search local repo" would give less surprising result. I have to give it a bit more thought.

nadako:
> For, haxelib remove haxelib i think it is reasonable to require the use of --safe option.

You mean, just exit with an error message (suggesting `--safe`) if we try to `remove haxelib` from within installed `haxelib`?

nadako:
UPD: added one more item we need to make sure is handled correctly

> Check if it works correctly when haxelib repo path is not configured (no /etc/.haxelib or ~/.haxelib file, no HAXELIB_PATH env var)

ibilon:
(haxelib remove --safe)
Not sure the safe is required, remove is already blocked against the set version, which would be running.
How far goes the windows directory lock?

(haxe upgrade doesn't upgrade haxelib)
For the haxe upgrade that would indeed not upgrade haxelib anymore (beside the safe version),
but I think that's ok since it'll be notified,
[...]

PR #295, opened 2016-03-12
Only use haxelib package globally

ibilon:
4th point of #293

this prevent haxelib to be use/load/updated/installed/run in a local repository


nadako:
Let me do that, I was in-progress, but I got distracted by @RobDangerous live stream https://www.twitch.tv/robdangerous :)

nadako:
Actually, I don't like neither my changes nor yours (they are very similar). Haxelib code is ugly enough already :)

How about we not restrict project-local haxelib installations, but only use one from the global repository and point user that he needs to add `--global` to update haxelib.

ibilon:
Special cases are never pretty,
as long as we don't load it and if we modify the notification it should be ok.

nadako:
I did that in b88a81bd0649a235f3fd10e68a189ab885069c0e.
We can revisit this when we'll be refactoring stuff, but this should be sufficient for now.

PR #298, opened 2016-03-12
Complete usage documentation

ibilon:
Now contains all the commands, flags and parameters (unless there's some I don't know about)


nadako:
Nice, thanks!

PR #302, opened 2016-03-13
Replace --safe flag with --system

ibilon:
#293


nadako:
I still don't know if `system` is really a better alternative to `safe`, but oh well...

PR #327, opened 2016-05-29
New layout / navigation

markknol:
This would be the last site with the "old" navigation

To have the menu the same as the rest, I needed downgraded to same bootstrap version as haxe org to keep styles the same. This has some changes on the layout (especially the buttons) but in general its not that bad.

I also did some extra stuff to make it a bit more nice.
- updated homepage layout. It still looking the same but with a bit more content.
- footer at bottom using flex layout
- updated tag browsing; keep tags on left side. unfortunately it still has the tagcloud (which I dislike) but I don't had inspiration yet to show something cooler there. Also tags look the same as on the cookbook. 

---

**Important** 

In the preview you see my local haxelib setup, i dropped random names and libraries in my database, can we put this into a staging with the current actual data? I first want to know sure if everything works before we actually put it live.
@andyli is that possible?

---

**Preview**

haxelib-new


markknol:
Arrgh what am i doing wrong, I see so much irrelevant commits ?

jdonaldson:
this looks great!   

FWIW, you need to rebase if you want to avoid a merge commit.  However, be careful not to do that if you've pushed changes somewhere publicly.  It will change the ordering of commits (essentially rewriting history).  https://git-scm.com/docs/git-rebase

andyli:
I dont think there is a staging site for haxelib at the moment, but i can
look into setting up one once i come back to Hong Kong next Monday.

markknol:
That would be great @andyli Please let me know when we can use this.

Simn:
What's the status here? Would be nice to finalize this.

nadako:
Yes please!

andyli:
I've set up an environment for the development branch. The url is http://development-lib-haxe-org.eu-west-1.elasticbeanstalk.com/. It is using the production database, so don't do funny thing with it (although we have auto daily database backup).

I will document this properly tmr. It will probably go to a Github wiki page.

markknol:
Great. I really want a staging environment to test with actual data. I cannot just put it live without proper testing, I dont know the project that well to be sure about everything.

markknol:
ping @andyli can you check this? it would be nice to have this live before/when Haxe 3.3 is release, soon.

andyli:
Sorry for the long delay... I've merge it and it will be live on http://staging-lib.haxe.org/ soon. Please check if everything is okay.

markknol:
Looks nice! Thanks

I do dislike the tags cloud http://staging-lib.haxe.org/t/ but I dont have a better idea how to improve/hide that.

Issue #238, opened 2015-08-18
support for git dependencies


clemos:
It doesn't seem possible to have git dependencies currently : 
```json
"dependencies": {
    	"hxnodejs": "https://github.com/HaxeFoundation/hxnodejs.git"
}
```
The above seems to fail. Or maybe there is another syntax ?
If not, then I think it would be nice to support git dependencies, pretty much like `npm` does.

nadako:
note that it would be great to also be able to specify commit number/branch/tag (like npm does I think)

clemos:
Yes, it's like `npm install git+ssh://git@github.com:npm/npm.git#v1.0.27`
see https://docs.npmjs.com/cli/install

jv-PintoBobcat:
+1 for this

nadako:
As mentioned in https://github.com/HaxeFoundation/haxelib/pull/135#issuecomment-163228477, I think we should adopt NPM dependency format, see https://docs.npmjs.com/files/package.json#dependencies.

ncannasse:
I'm fine with following anything that is quite standardized as NPM seems to be, as long as we keep backward compatibility with our previous format, so users can continue downloading and submitting libraries using the old format.

fzzr-:
1. In this case haxelib.json must be exists in root of repo;
2. What about another (potential) supported systems such as Mercurial or Subversion? Are we have to detect what use by url :) ? No.

porfirioribeiro:
Maybe something like Cargo from rust lang?
http://doc.crates.io/manifest.html#the-[dependencies]-section
It accepts git repo's as a depedency. The Cargo.toml is either used on Lib's or app's to specify options including dependencies.

nadako:
That is interesting stuff, thanks!

fzzr-:
I like the Cargo impl. but this is new format and will change the our current format:
From `{"libname" : "ver"}` to `{"libname" : {"version":"", "git/hg/svn":"", "branch/tag/rev":""} }`.

But we can really preserve backward compatibility. We can easy support old(current) and new formats - detect format via Reflection and use the special provider (parser) for it _(Or EitherType)_.

fzzr-:
Another one variant - namespace-like:
```json
"dependencies": {
        "libFromHaxelibA": "",
        "libFromHaxelibB": "10.2.4",
        "libFromGit": "git::https://github.com/HaxeFoundation/libFromGit.git",
[...]

PR #344, opened 2016-08-31
Git dependencies

st3veV:
Git dependencies behavior as discussed in #238 :
- it's now possible to call `haxelib install haxelib.json` and dependencies from the json will be installed
- dependencies can have version in form: `git:<url>#commit-ish`... if so git dependency will be installed, this is possible in both json and hxml definitions
- if `commit-ish` is defined and/or different than one already installed from same json/hxml (ie. because of dependency of dependency) user is prompted to confirm overwriting existing git version


andyli:
Thanks. Would you take a look at the failed test and try to fix it?
```
Command: haxe [client_tests.hxml]
Class: tests.TestSemVer ...
Class: tests.TestData .E.......
Class: tests.TestRemoveSymlinks .
Class: tests.TestRemoveSymlinksBroken .
Class: tests.TestHg ...........
Class: tests.TestGit ...........
Class: tests.TestVcsNotFound ....
Class: tests.TestInstall ..
* tests.TestData::testReadDataWithoutCheck()
ERR: exception thrown : Invalid field access : length
Called from StringTools::startsWith line 142
Called from haxelib._Data.Dependencies_Impl_::toArray line 83
Called from haxelib.Data::readData line 339
Called from tests.TestData::testReadDataWithoutCheck line 237
Called from a C function
Called from haxe.unit.TestRunner::runCase line 115
FAILED 42 tests, 1 failed, 41 success
Command exited with 1 in 55s: haxe [client_tests.hxml]
```

st3veV:
Fixed.

ibilon:
Hey, thanks for the contribution.

Could you add a check that vcs dependencies aren't allow when submitting to haxelib?

> if commit-ish is defined and/or different than one already installed from same json/hxml (ie. because of dependency of dependency) user is prompted to confirm overwriting existing git version

Not sure I fully get what that means, you can't have two branch of the same repository used as deps by different libs/project?

st3veV:
Yep, the check makes sense. I'll look at it.

>Not sure I fully get what that means, you can't have two branch of the same repository used as deps by different libs/project?

Correct. In the discussion for #238 it was decided to not bother with this and support just one because it already brings quite a lot of value (https://github.com/HaxeFoundation/haxelib/issues/238#issuecomment-223231264 and https://github.com/HaxeFoundation/haxelib/issues/238#issuecomment-224905389).

st3veV:
So I did some digging through the code and testing and there already is a "check"... submit fails with ie.: `Error: git:https://github.com/fzzr-/hx.signal.git is not a valid version string` because the JSON is being validated. In my opinion the error descriptive enough and submit is failing so it behaves as expected but I'm open to different opinions.

ibilon:
The error message could be improved but it's still descriptive enough.

@nadako I haven't tried it yet but the design and code are good, what do you think?

nadako:
lgtm, ship it!

PR #346, opened 2016-09-03
Website: only update a project shown version if it's a higher semver

ibilon:
`versionObj` is used to store the current version of a lib
ie. the one shown when looking at it http://lib.haxe.org/p/haxelib_client

It's not the same as the one selected by the client as latest which uses semver.

Currently if you upload a patch to an old version it'll overwrite `versionObj`.

### DIFF
--- a/src/haxelib/server/Repo.hx
+++ b/src/haxelib/server/Repo.hx
@@ -315,7 +315,11 @@ class Repo implements SiteApi {
 				v.documentation = doc;
 				v.insert();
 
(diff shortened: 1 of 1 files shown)


nadako:
This makes me wonder, maybe we should get rid of `versionObj` and `curversion`, and make it consistently search for the latests one...

andyli:
Currently CI is failed due to this change. I can reproduce the test failure locally. Investigating.

PR #347, opened 2016-09-03
Some project from the DB are null, don't try to construct object for …

### LINKED ISSUE #342: Viewing the cross tag errors
When browsing to http://lib.haxe.org/t/cross, I get an error message:
`website.controller.HomeController.tag()`

All other tags (including empty/fake ones) work normally.

ibilon:
…them

Should fix #342 http://lib.haxe.org/t/cross/ fails with `Invalid field access : name`
but http://lib.haxe.org/t/cross.json/ works and contains some `null`


nadako:
Hmm, @andyli is that because we removed some haxelibs from the projects table, but forgot to remove relevant links from tags? I'll merge this because it'll wont hurt anyway, but there's some invalid records in tags.

PR #355, opened 2016-09-15
Git dependencies fix

st3veV:
Library and branch must be stored after the first installation as well otherwise it'll be updated when another library depends on git version without specified commit-ish.

Example:
- Clean lib folder or `newrepo`
- Top level git dependency with specific commit hash
- Another top level git dependency depending on the previous one but without specified commit-ish
- First one is cloned properly
- When second one's dependencies are being cloned it should not do anything with the existing git repo of the first one (old behavior was that it updated it which shouldn't happen as specified version/branch must always precede any "don't care" version/branch)


nadako:
Thanks!

PR #365, opened 2017-01-17
Caching proxy server

st3veV:
Haxelib server version which can automatically cache requested libraries and forward calls to a "parent" server when it's missing something.

The purpose of this server is to be running in a company which has closed internal libraries that should not be submitted to the public repository and also want's to have internal copy of needed dependencies in case of something happens to the public repo (outage, purge, npm-leftpad-like problem)

Answers to some questions you might have to save your time.
Q: Why is it separate file and not part of the original server?
A: I wanted to keep it as lightweight as possible and the original doesn't contain file handling (that is done by the ufront web part and is not part of the API). The proxy server only needs nekotools server to run.

Q: Why so weird http request parsing for submitted files?
A: Nekotools server doesn't support multipart request parsing and this solution (even though simplistic and ugly) works so far.

Q: Database cache of users is completely unsafe and there are just dummy data, why?
A: There have to be some values to keep the database valid and consistent, because caching only files is not enough, infos has to be cached as well because the client is checking them before requesting any file.
Usually there will be only two entities adding something to the database:
    1. CI server submitting the internal libraries - here the security is the same as the standard version
    2. The proxy server itself when it's caching new dependency where proper authentication can't be done because it's not submitted by the original author.

Comments and suggestions are very welcome.

EDIT:
nekotools server turned out to be unusable so I replaced it with apache using simplified rewrite rules

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,10 +12,11 @@ matrix:
       env:
         - TRAVIS_HAXE_VERSION=development
         - DEPLOY=1
+      services: [mysql]
       before_install:
         - sudo add-apt-repository ppa:haxe/snapshots -y
         - sudo apt-get update
-        - sudo apt-get install haxe=1:3.4* neko-dev libapache2-mod-neko mysql-server mysql-client -y
+        - sudo apt-get install haxe=1:3.4* neko-dev libapache2-mod-neko -y
         - mkdir ~/haxelib && haxelib setup ~/haxelib
         - sudo usermod -a -G travis www-data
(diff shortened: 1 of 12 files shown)


st3veV:
Clearly I messed up the branch and now there are tons of duplicate commits. If there is a decision to accept this I'll create a new pull request with everything squashed together. We can still use this as work-in-progress.

cmandlbaur:
Has there been any update on this?

Simn:
Closing due to inactivity.

Issue #380, opened 2017-04-05
Vulnerability with "releasenote" on website


starburst997:
Seems like there is a possibility to inject html (and perhaps javascript?) to the website via the "releasenote"

![image](https://cloud.githubusercontent.com/assets/77518/24708697/a9350c96-19e5-11e7-84ca-25efe2a9c711.png)

Just saw this on the front page and thought it was worth mentioning, did not try anything but it would probably be very easy to create a "fake" link to trick user or if it's possible to inject javascript to replace all links, etc.

The "releasenote" of "js-virtual-dom"
```
"releasenote": "some parameters added for <input type=\"range\" />",
```

starburst997:
Also visible in the "versions" page of the project

![image](https://cloud.githubusercontent.com/assets/77518/24709205/2787e43c-19e7-11e7-8792-c062fcf7b20d.png)

PR #381, opened 2017-04-05
Fix for #380

starburst997:
It is indeed possible to write Javascript code inside "releasenote" and get them executed on the website. This fix it. But there might be some other area like "description" / "name"  / "contributors", even license that might be vulnerable???

A simple
```
"releasenote": "Just a normal message <script>console.log('Some evil code');</script>",
```

Inside a haxelib.json is enough.

### DIFF
--- a/src/website/controller/HomeController.hx
+++ b/src/website/controller/HomeController.hx
@@ -106,6 +106,7 @@ class HomeController extends Controller {
 			title: 'Haxelib Tags',
 			description: 'The 50 most popular tags for projects on Haxelib, sorted by the number of projects',
 			tags: tagList,
+			escape: function(str:String) return StringTools.htmlEscape(str, true),
 			tagCloud: tagCloud,
 		});
 	}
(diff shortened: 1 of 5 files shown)


Simn:
This is important, injection like that is very embarrassing in 2017...

starburst997:
I'm currently trying to test every possible fields on my dev haxelib and see if there is other vulnerability like this elsewhere.

starburst997:
When using "register" on haxelib, you can get injection on the website with the "Full name"
```
Fullname : Name<script>console.log('full');</script>
```

fullofcaffeine:
Can't we have automatic sanitization on the view layer?

starburst997:
Yeah, I didn't want to go too deep inside the code, but it seems like this is worst than I thought and this would mean htmlEscape everywhere, so might as well get that on the view layer.

starburst997:
Tags are vulnerable to this also

starburst997:
"url" and "description" also

fullofcaffeine:
@starburst997 >EDIT: Btw, what's the template rendering engine used here?

starburst997:
Erazor
https://github.com/ufront/erazor

fullofcaffeine:
@starburst997 I can see an whitelisting approach where all interpolated values are escaped by default. Whatever needs to render actual HTML can then be added to a whitelist somehow (maybe by wrapping it in a call to `htmlSafe()` or something).

I think that would need to be hacked at the erazor level. I don't have time right now (using my lunch time now to answer to this thread) otherwise I'd look. But might be able to do that over the weekend.

Simn:
Thanks for this!

Issue #351, opened 2016-09-04
Haxelib creates ~ folder on Ubuntu


PeyTy:
Tested on 3.2.1

``` php
peyty@peyty-U16041:~$ haxelib setup
Please enter haxelib repository path with write access
Hit enter for default (/usr/share/haxelib)
Path : ~/.haxelib           
haxelib repository is now /home/peyty/~/.haxelib/
```

It _really_ creates `/home/peyty/~/.haxelib/`:

![image](https://cloud.githubusercontent.com/assets/3642643/18222487/598da780-71a1-11e6-9365-2ce434956f3d.png)

---

Expected behavior: ~ is same as /home/peyty, so I expect /home/peyty/.name folder to be created.
From @ibilon 

> you can't do ~/.haxelib/ (even if ~ was correctly detected) because haxelib use a ~/.haxelib file to store data

Okay then, at least ~/.haxelib should show error or something, and ~/.haxe path must work I think.

``` php
peyty@peyty-U16041:~$ ls ~
~                 mm.cfg        Документы    Музыка         Шаблоны
Dropbox           stencylworks  Загрузки     Общедоступные
examples.desktop  Видео         Изображения  Рабочий стол
peyty@peyty-U16041:~$ ls /home/peyty/
~                 mm.cfg        Документы    Музыка         Шаблоны
Dropbox           stencylworks  Загрузки     Общедоступные
examples.desktop  Видео         Изображения  Рабочий стол
```

L4D15:
This is happening to me too in macOS.

mlms13:
If your instincts tell you to `rm ~`, don't do it. I did it and it wasn't pretty. :P

paulorobertocruz:
Is this still a problem? I will try working on this.

tobil4sk:
This was resolved in #382

PR #382, opened 2017-04-29
Fix for issue #351

paulorobertocruz:
When calling haxelib setup ~/path/to/haxelib '~' will be translated to the user
home dir.


Aurel300:
Err … This should be handled by the shell, never a program itself. If you do `echo ~` in bash, it is bash that expands the tilde, not `echo`. This should be reverted @RealyUniqueName

ibilon:
@Aurel300 for the command line version yes, but that code also deal when using haxelib interactively.
```
haxelib setup
Path : ~/path/to/haxelib
```

Aurel300:
Oh right, I see. Would be nice if we had a `wordexp` equivalent, but I guess this will have to do for now!

PR #384, opened 2017-05-18
Add wildcard version support

jgranick:
This is the start of a set of updates to add wildcard version support to haxelib

 1.) This enables support for `haxelib install mylib 5.*` or `haxelib path mylib:1.0.*`
 2.) This enables `haxelib path mylib:1.0.1` even if a newer "dev" directory is set

`haxelib path` is unpredictable, because the "dev" directory always overrides the set version. That is valuable for getting the current library version, but this is troublesome if you are trying to retrieve the path of an older.

These changes reads "haxelib.json" of dev paths in order to determine what version it is.

Imagine that there is a library called "mylib", with a development directory (version 1.2.0) set:

```bash
mylib: 1.0.0 1.0.1 1.1.0 [dev:C:\mylib]
```

Currently haxelib will return the following:

```bash
> haxelib path mylib
C:\mylib
-D mylib=1.2.0
> haxelib path mylib:1.1.0
C:\mylib
-D mylib=1.1.0
> haxelib path mylib:1.0.0
C:\mylib
-D mylib=1.0.0
```

This is very confusing, and prevents us from getting older paths. These changes enable the following output

```bash
> haxelib path mylib
C:\mylib
-D mylib=1.2.0
> haxelib path mylib:1.1.0
C:\HaxeToolkit\haxe\lib\mylib\1,1,0
-D mylib=1.1.0
> haxelib path mylib:1.0.0
C:\HaxeToolkit\haxe\lib\mylib\1,0,0
-D mylib=1.0.0
> haxelib path mylib:1.*
C:\mylib
-D mylib=1.2.0
> haxelib path mylib:1.1.*
C:\HaxeToolkit\haxe\lib\mylib\1,1,0
-D mylib=1.1.0
> haxelib path mylib:1.0.*
C:\HaxeToolkit\haxe\lib\mylib\1,0,0
-D mylib=1.0.0
```

This should enable use of a "dev" directory matching an older version number, and developers who always use "dev" for current library versions, but may need to build a project using an older library version as well. This only affects projects that request an exact version, instead of the current version

The goal of wildcard version numbers is to allow looser dependencies. According to SemVer, both minor and patch releases should remain backward compatible. A developer may choose to require "mylib" version "2.*" in this way, without updating each time there is a new minor or patch release, or without supporting any installed version

### DIFF
--- a/src/haxelib/client/Main.hx
+++ b/src/haxelib/client/Main.hx
@@ -652,16 +652,21 @@ class Main {
 		if( inf.versions.length == 0 )
 			throw "The library "+inf.name+" has not yet released a version";
 		var version = if( reqversion != null ) reqversion else inf.getLatest();
-		var found = false;
+		var matches = [];
+		var best = null;
 		for( v in inf.versions )
-			if( v.name == version ) {
-				found = true;
-				break;
+			if( matchVersion(version,v.name) ) {
+				matches.push(v.name);
 			}
-		if( !found )
+		for( match in matches ) {
+			if (best == null || match > best) {
+				best = match;
+			}
+		}
+		if( best == null )
 			throw "No such version "+version+" for library "+inf.name;
 
-		return version;
+		return best;
 	}
 
 	function installFromHxml( rep:String, path:String ) {
--- a/src/haxelib/client/Main.hx
+++ b/src/haxelib/client/Main.hx
@@ -1086,16 +1086,23 @@ class Main {
 				continue;
 			v = Data.unsafe(v);
 			var semver = try SemVer.ofString(v) catch (_:Dynamic) null;
+			if ( semver == null ) {
+				var json = try File.getContent(dir+"/"+v+"/"+Data.JSON) catch( e : Dynamic ) null;
+				if ( json != null ) {
+					var inf = Data.readData(json, false);
+					semver = try SemVer.ofString(inf.version) catch (_:Dynamic) null;
+				}
+			}
 			if (semver != null && matchVersion(version, semver))
-				matches.push(semver);
+				matches.push({ dir: v, ver: semver });
 		}
-		var best = null;
+		var best:Dynamic = null;
 		for( match in matches ) {
-			if (best == null || match > best) {
+			if (best == null || match.ver > best.ver) {
 				best = match;
 			}
 		}
-		return if (best != null) dir + "/" + Data.safe(best) else null;
+		return if (best != null) dir + "/" + Data.safe(best.dir) else null;
 	}
 
 	function list() {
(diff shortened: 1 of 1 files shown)


ibilon:
Haxe uses the path command, so if you change it like that dev loses its overriding capacity no?

jgranick:
"dev" will still override if you do not ask for a specific version, or the version matches.

In my example above, `haxelib path mylib`, `haxelib path mylib:1.2.0`, `haxelib path mylib:1.*` or `haxelib path mylib:1.2.*` return the dev directory. Actually, the changes also fixes `haxelib path mylib:1.2.0`, which returns an error without the changes (since no non-dev library exists by that version)

ibilon:
> "dev" will still override if you do not ask for a specific version

It's an override because it returns itself even when you ask for a specific version.
Otherwise it's just a set to a folder, which would be useful, but doesn't encompass the use case of dev.

jgranick:
I am open to suggestions

An example use-case is a system using development versions of the latest Lime and OpenFL

A user on this system wants to build a project using old versions. There is currently no way to resolve paths to the other versions installed without browsing the contents of the haxelib repository folder in Lime tools (which is not a good path to take).

A development directory can only ever be one version. A system that wants new and old versions of the same library must reasonably be able to distinguish between the two

ibilon:
I think dev is currently misused, because of the lack of proper support for "local version".
Its purpose is to override the library to allow overriding a project/build system without changing anything.

Personally I think it is working fine, and what we need is support for the use of a path as version instead.

Two possibilities:

* having directly the path as value in set/path `haxelib set mylib /some/path/`
* introducing a local command `haxelib add-local mylib fromgit /some/path/` which would make it available.
  ```
> haxelib list mylib
mylib: 1.0.0 [fromgit]
```

jgranick:
Alright, here is an idea for how we could handle it

 * `haxelib install actuate`
 * `haxelib install build.hxml`
 * `haxelib install package.zip`
 * `haxelib install path/to/actuate`

I notice that `haxelib install` was expanded to support HXML files. I think we could overload the `install` command to support ZIP package files (like `haxelib local`) as well as directory paths that contain a "haxelib.json" file.

After install, the version is set based upon the version found initially in the JSON file.

The current dev behavior can remain the same, but this would allow users to install a local directory of a library, preventing the strict override behavior. 

```bash
> haxelib list mylib
mylib: 1.0.0 [1.0.1:C:\Development\Haxe\mylib-1.0.1]

What do you think?

ibilon:
You can already do `haxelib install package.zip` ;)

> After install, the version is set based upon the version found initially in the JSON file.

That's the issue with local directory as library, is that their version can change anytime.
And it could also be the same as an already installed version.

Either we authorize path only as the set version, in which case it'll only be used when no specific version is asked.
Or we use the version from their haxelib.json, but in that case there are more issues to figure out.

And I think the first case should enough for the development version use case?
Which should be the only one requiring to use a path as library (the only one I can think of).

jgranick:
I'll work more on local directory support, and circle back afterward :smile:

PR #385, opened 2017-05-21
support multiple repositories on HAXELIB_PATH

volth:
support multiple repositories on HAXELIB_PATH (akin to PERL5LIB and CLASSPATH)

following nixos discussion https://github.com/NixOS/nixpkgs/pull/25953

### DIFF
--- a/src/haxelib/client/Main.hx
+++ b/src/haxelib/client/Main.hx
@@ -145,6 +145,7 @@ class Main {
 		apiVersion : "3.0",
 	};
 	static var IS_WINDOWS = (Sys.systemName() == "Windows");
+	static var PATH_SEPARATOR = if (IS_WINDOWS) ";" else ":";
 
 	var argcur : Int;
 	var args : Array<String>;
@@ -360,11 +361,15 @@ class Main {
 		}
 
 		if (!isHaxelibRun && !settings.system) {
-			var rep = try getGlobalRepository() catch (_:Dynamic) null;
-			if (rep != null && FileSystem.exists(rep + HAXELIB_LIBNAME)) {
-				argcur = 0; // send all arguments
-				doRun(rep, HAXELIB_LIBNAME, null);
-				return;
+			var reps = try getGlobalRepositories() catch (_:Dynamic) null;
+			if (reps != null) {
+				for (rep in reps) {
+					if (FileSystem.exists(rep + HAXELIB_LIBNAME)) {
+						argcur = 0; // send all arguments
+						doRun([rep], HAXELIB_LIBNAME, null);
+						return;
+					}
+				}
 			}
 		}
 
@@ -611,8 +616,12 @@ class Main {
 		return password;
 	}
 
+	function getWritableRepository(reps: Array<String>):String {
+		return reps[0];
+	}
+
 	function install() {
-		var rep = getRepository();
+		var rep = getWritableRepository(getRepositories());
 
 		var prj = param("Library name or hxml file:");
 
@@ -788,7 +797,7 @@ class Main {
 		// check if exists already
 		if( FileSystem.exists(rep+Data.safe(project)+"/"+Data.safe(version)) ) {
 			print("You already have "+project+" version "+version+" installed");
-			setCurrent(rep,project,version,true);
+			setCurrent([rep],project,version,true);
 			return;
 		}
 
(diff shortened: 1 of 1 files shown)


src/haxelib/client/Main.hx
```diff
@@ -954,27 +963,35 @@ class Main {
...
-			rep = getWindowsDefaultGlobalRepositoryPath();
+			repSystem = getWindowsDefaultGlobalRepositoryPath();
 			if (create)
-				try safeDir(rep) catch(e:Dynamic) throw 'Error accessing Haxelib repository: $e';
+				try safeDir(repSystem) catch(e:Dynamic) throw 'Error accessing Haxelib repository: $e';
+		}
+		if (repEnv == null && repUser == null && repSystem == null)
+			throw "This is the first time you are runing haxelib. Please run `haxelib setup` first";
+		var rep: String = "";
+		if (repUser != null)
```
volth: The first in the ```reps``` list will be writable repo for ```haxelib install```, so the order matters and that's why ```repUser``` is the first to try - it is the setting of ```haxelib setup```.

ibilon:
Hi, thanks for contributing.

If I understand correctly the use case is to allow having multiple storage directories for the libraries?
Mainly to have one global system wide directory and a user specific I guess?

src/haxelib/client/Main.hx
```diff
@@ -145,6 +145,7 @@ class Main {
 		apiVersion : "3.0",
 	};
 	static var IS_WINDOWS = (Sys.systemName() == "Windows");
+	static var PATH_SEPARATOR = if (IS_WINDOWS) ";" else ":";
```
ibilon: I'm not sure if it's worth having os specific separator. Any reasons?

volth: It is modelled after PERL5LIB and Java's CLASSPATH.

```:``` is a quite common symbol on Windows and cannot be used as the delimiter (for example ```C:\Program Files\Haxe\haxe.exe``` should not be split to ```C``` and ```\Program Files\Haxe\haxe.exe```).

Using ```;``` on Linux is arguable, ```:``` might be a sort of tradition. I do not know its history.

**Comment by volth** [comment]
Use case is to have HAXELIB_PATH set to something like ```/nix/store/a81b6s16mznpk8lwxqar7cnnc0kkylmp-haxe-3.4.2/lib/haxe:/nix/store/5zmk2yjim1gckpd11vb992vlk9g74ii7-hxjava-3.2.0/lib/haxe:/nix/store/59amv9blqhcw0zyb09zyv45q09nq66s9-hxcs-3.4.0/lib/haxe:/nix/store/dzy7lkssaz
3a7fcw4b8as2anfx69h5bc-hxcpp-3.4.64/lib/haxe:/nix/store/j28yj4kml4rhrwbs594x9z3r0wfcjr42-lime-5.0.0/lib/haxe:/nix/store/s0b20dc1dl5ybfmd5758x5ma0y0i9iz1-msgpack-haxe-1.15.1/lib/haxe:/nix/store/b87pxzmgyjd6hxhgdl6mqs3ch1rfy5w3-fo
rmat-3.3.0/lib/haxe:/nix/store/phgfaxw1g7fhs2srfjryrarwjsna0j5q-openfl-5.1.0/lib/haxe:/nix/store/3rdzxk05qck3bji4gsl9rn2zxk1hfb96-openfl-samples-4.9.0/lib/haxe:/nix/store/b6liq38jphg56hsgm2pljq35b0yjyq58-flixel-4.2.1/lib/haxe:/n
ix/store/x8cg4rx1hcyk8pl065k9zvi7hgv07fx4-actuate-1.8.7/lib/haxe```

The path like this is created by a [setup hook](https://github.com/volth/nixpkgs/blob/23def3e5f55d55eeb5102fecdd576b87af4eb281/pkgs/development/compilers/haxe/setup-hook.sh) by combining pathes of the installed libraries (or the libraries mentioned as ```buildInputs``` of another package). That is how Nix works with Perl, Python, Java, ...

File: src/haxelib/client/Main.hx (hunk 2 above)
Comment (back2dos) [inline src/haxelib/client/Main.hx]: It's the platform specific delimiter that is used in `PATH` (among other things), see https://nodejs.org/api/path.html#path_path_delimiter.

**Comment by volth** [comment]
CI failed on Windows, I will investigate it later today

File: src/haxelib/client/Main.hx (hunk 2 above)
Comment (ibilon) [inline src/haxelib/client/Main.hx]: Right, forgot that you couldn't use `:` on windows.
In that case following perl/java/node makes sense.

**Comment by volth** [comment]
The failure seems to happen in old ```haxelib.exe``` which is installed by Chocolatey.
It does not split the path

**Comment by volth** [comment]
[Here](https://github.com/volth/nixpkgs/blob/04668f031daac8ce4cf1938254b3763299de0f74/pkgs/development/compilers/haxe/haxe-packages.nix#L138) is an use case example.

All ```/bin/``` directories (if any) of packages mentioned in ```buildInputs``` form ```$PATH```.
Similarly ```/lib/perl5/site_perl/``` form ```$PERL5LIB```, and ```/lib/haxe/``` form ```$HAXELIB_PATH```

The bash script below is able to run ```haxe``` (it is on ```$PATH```), and haxe is able to compile C++ example (```hxcpp``` is on ```$HAXELIB_PATH```) and Java example (```hxjava``` is on ```$HAXELIB_PATH``` and ```jdk``` is on ```$PATH``` transitively as ```propagatedBuildInputs``` of ```hxjava```). Unfortunately, the Flash example is broken, it fails with "SecurityError: Error #2017: Only trusted local files may cause the Flash Player to exit.", so I commented it out.

That's basically how Nix build things.

File: src/haxelib/client/Main.hx [hunk 3]
```diff
@@ -611,8 +616,12 @@ class Main {
 		return password;
 	}
 
+	function getWritableRepository(reps: Array<String>):String {
+		return reps[0];
```
andyli: When `haxelib setup` is never run yet, e.g. when haxe is just installed, this may point to the system repo (`HAXELIB_PATH`)?

volth: Yes, ```haxelib update``` is also questionable, it may end up using the read-only repo where previous version is installed.

volth:
There is one issue with ```lime``` having its own copy of ```haxelib``` https://github.com/openfl/lime/tree/develop/tools/haxelib so it should be patched in sync (or they first)

PR #386, opened 2017-05-23
Add support for custom local directories and wildcard versions

jgranick:
This adds support for two additional features:

## Custom Local Directories

`haxelib dev` enables the use of a custom local directory, but forces an "override always" behavior that can be undesirable. This adds support for custom named directories, with the default being called "local". The current syntax is as follows:

```bash
haxelib install path/to/mylib
haxelib install path/to/mylib custom_name
```

`haxelib install path/to/mylib` will install "local" with a path to "path/to/mylib", while adding the optional "custom_name" argument installs it as "custom_name"

> 1.0.0 1.0.1 [local:path/to/mylib]

You should be able to remove, set and path similar to other types of versions

## Wildcard Versions

There is also added support for wildcard versioning:

```bash
haxelib version mylib
haxelib version mylib:1.0.0
haxelib version mylib:1.*
haxelib version mylib:1.0*
haxelib version mylib:1.0.0-*
haxelib version mylib:1.0.0-beta*
```

This should work for "install" and "path", there may be some commands (such as "set") that still need to have support added

I would be happy to discuss these features, and how they can be polished to fit in best

### DIFF
--- a/src/haxelib/client/Main.hx
+++ b/src/haxelib/client/Main.hx
@@ -652,16 +652,21 @@ class Main {
 		if( inf.versions.length == 0 )
 			throw "The library "+inf.name+" has not yet released a version";
 		var version = if( reqversion != null ) reqversion else inf.getLatest();
-		var found = false;
+		var matches = [];
+		var best = null;
 		for( v in inf.versions )
-			if( v.name == version ) {
-				found = true;
-				break;
+			if( matchVersion(version,v.name) ) {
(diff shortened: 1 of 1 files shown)


tobil4sk:
Closing. "Local" versions are no longer needed as `dev` no longer overrides when specific versions passed into `path` (7bd9263) or `run` (63baf8a), and installing locally from zip is also available. The remaining changes would probably require a new PR.

PR #394, opened 2017-09-26
Minor typos in README.MD

wiggin77:


Simn:
"We recommand" is pretty funny... but thanks anyway!

PR #399, opened 2017-11-05
Allow symlinks for haxe/lib directory

sebthom:
This change prevents the error `Error: Error accessing Haxelib repository: A file is preventing C:\apps\dev\haxe\3.4.4\lib to be created
` that is raised when the lib directory within a Haxe installation is a symlink. 

Symlinking the lib directory is useful if multiple installations of Haxe exist and this can share the local lib repository.

Creating symlinks on Windows can be done using `mklink /D c:\apps\dev\haxe-3.4.4\lib c:\apps\dev\haxe-shared-lib`.
On Windows using symlinks (mklink /D) instead of hardlinks (mklink /H)  is preferable as it does not require administrative rights and thus can be automated in a non-elevated batch script.

### DIFF
--- a/src/haxelib/client/FsUtils.hx
+++ b/src/haxelib/client/FsUtils.hx
@@ -1,5 +1,5 @@
 /*
- * Copyright (C)2005-2016 Haxe Foundation
+ * Copyright (C)2005-2017 Haxe Foundation
  *
  * Permission is hereby granted, free of charge, to any person obtaining a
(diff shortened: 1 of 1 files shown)


sebthom:
@andyli Hi Andy, can you please review this. Thanks!

andyli:
Looks good to me. Thanks!

sebthom:
Nice. Thanks for the fast merge!

PR #408, opened 2018-02-08
Adds `haxelib libpath` - Closes #407

### LINKED ISSUE #407: `haxelib libpath` to return root path of a library
~Currently it seems that `haxelib path` returns a lot of unformatted extra-information, instead of just the path. It would be great if the extra information was hidden behind a flag, or a flag could be provided that would only return the path of the library - as the command suggests.~

`haxelib path` returns the paths of libraries' source files + compiler flags. It would be great to have a command where just the `root path` of a library would be returned.

This way `haxelib libpath` can be used to programmatically work with library locations - e.g. to detect current [git commit of libraries in use at build time](https://code.haxe.org/category/macros/add-git-commit-hash-in-build.html), etc.

Desired output : 
```bash
$ haxelib libpath openfl-samples
/usr/local/lib/haxe/lib/openfl-samples/4,0,0/
```

```bash
$ haxelib libpath openfl
/somewhere/unexpected/openfl/
```

```bash
$ haxelib libpath tink_macro
/usr/local/lib/haxe/lib/tink_macro/0,15,4/
```

mastef:
Adds `haxelib libpath` - Closes #407

### DIFF
--- a/test/RunCi.hx
+++ b/test/RunCi.hx
@@ -471,6 +471,9 @@ Listen 2000
 			compileClient();
 			testClient();
 			compileServer();
+		#end
+		// buddy is only compatiable with haxe 3.4+
+		#if (haxe_ver >= 3.4)
 			testServer();
 		#end
 


EricBishton:
I think you need to bump the version, too.  So we can programatically know whether the command is available on the user's system.

src/haxelib/client/Main.hx
```diff
@@ -1344,6 +1345,25 @@ class Main {
 		}
 	}
 
+	function pathonly() {
```
EricBishton: OK, this is cheating. 😃 In the interest of not duplicating bugs, do you think that we can combine the contents of the two commands into one shared routine?

mastef: OK @EricBishton @kLabz I didn't want to do it at first, since it'd require a `.bind` and would make reading the method harder. But seems other methods also use binds - so I've merged them now.

mastef:
@EricBishton bumped to 3.3.1

back2dos:
Uhm, I suppose the first question is: what do you mean by the "path" of a library? The location where it's on the file system? Because that's not what the first line of `haxelib path` will tell you. What it will tell you is the class path of the sources of the library. In most libraries, it tends to be `src` within the library, while in a few others it's indeed the library root, but on occasion it's somethin else like in promhx, where [it's `src/main`](https://github.com/jdonaldson/promhx/blob/master/haxelib.json#L5). Your enthusiasm is appreciated, but might I propose to first state your goal and then write code? ^^

mastef:
Hi @back2dos this PR follows a discussion on gitter, and the intent was stated in #407 : `or a flag could be provided that would only return the path of the library`.

Apologies if not everything has been stated here. You're right that the current usage of `path` is confusing - as it'd be confusing for anybody else reading the documentation. Your edge-case is perfectly valid critique and should be handled.

In this case it looks like `haxelib path` is more of a `haxelib srcpath`. And what I was looking for is a `haxelib libpath` - something you'd expect being returned when using a command called `haxelib path <lib>` with the vague description `give paths to libraries`.

mastef:
So I could re-do this and fix the documentation of `haxelib path <lib>` to `give paths to libraries' sources`, and add a command called `haxelib libpath <lib>` with `returns the root path of a library`.

back2dos:
Ok, sorry for the confusion. May I suggest to give slightly more context next time? In a year even those who followed the gitter discussion will have forgotten it, and then the PR will be as enigmatic to everyone, as it has been to me.

As for the rest: as Phillipe [stated](https://github.com/HaxeFoundation/haxelib/issues/407#issuecomment-364065496) it's a separate feature. For that reason alone, I'd not interlace it with `haxelib path`. Also, `haxelib path` is a very critical part of the client (it's what allows Haxe to resolve libraries) and therefore I think we should try to keep it as simple as possible.

What I would definitely advise is to make sure the logic is exposed for programmatic use, so that when compiling with `-lib haxelib_client` one can use it directly, rather than passing through the OS. It may be argued that calling `haxelib` is a better idea, because then you know the result is 100% consistent with haxelib, but on the other hand this means relying on the user having the latest haxelib installed (to actually support this command). In the near future, that's a highly unsafe assumption to make.

mastef:
Yes, well I stated it in the linked issue 😄

@back2dos I've updated the PR now though to be separate from `haxelib path`. I've undone any changes to the `path()` logic - and called the new command `haxelib libpath`.

src/haxelib/client/Main.hx
```diff
@@ -1306,7 +1307,7 @@ class Main {
 			if( p.project == prj ) {
 				if( p.version == version )
 					return;
-				throw "Library "+prj+" has two version included "+version+" and "+p.version;
+				throw "Library "+prj+" has two versions included : "+version+" and "+p.version;
```
kLabz: Nice catch for the plural :) There is no space before `:` though

src/haxelib/client/Main.hx
```diff
@@ -1344,6 +1345,16 @@ class Main {
 		}
 	}
 
+	function libpath( ) {
+		var rep = getRepository();
+		while( argcur < args.length ) {
+			var a = args[argcur++].split(":");
+			var results = new List();
+			checkRec(rep, a[0],a[1],results);
+			if( !results.isEmpty() ) Sys.println(results.first().dir);
```
kLabz: Why only `results.first()`?

back2dos: Well, I suppose the question is again what exactly you want. Assuming you really want just the one path (and not the dependencies), this is fine. But then looping the args is nonsense. Instead you should make sure there's exactly one argument left and process just that.

src/haxelib/client/Main.hx
mastef: @kLabz What do you mean with `There is no space before : though`? There's a space before and after the column ` : ` same as in [L1297](https://github.com/mastef/haxelib/blob/6e09f72a82c79b7da302ec567c9948d887cecfd5/src/haxelib/client/Main.hx#L1297) which is 13 lines above it. ( And [L918](https://github.com/mastef/haxelib/blob/6e09f72a82c79b7da302ec567c9948d887cecfd5/src/haxelib/client/Main.hx#L918) )

src/haxelib/client/Main.hx
mastef: `results.first()` will give you the library's directory, which `haxelib libpath` should return. Any other paths will be dependency paths.

The looping is replicating the behaviour from `haxelib path openfl lime` - if you run with `haxelib libpath openfl lime` it will return the paths for both libraries.

If you only run it with `haxelib libpath openfl` you will only get one path.

Since one `path` method works that way, I didn't want to make it confusing by changing the behaviour of the `libpath` method. The end-user doesn't have to supply multiple arguments.

mastef:
@back2dos I want to be very clear on what this command should accomplish : 

After we run `haxelib install openfl`
`haxelib libpath openfl` should return:
`/usr/local/lib/haxe/lib/openfl/7,0,0/`

After we run `haxelib git openfl https://github.com/openfl/openfl`
`haxelib libpath openfl` should return:
`/usr/local/lib/haxe/lib/openfl/git/`

After we run `haxelib dev openfl /projects/my/openfl`
`haxelib libpath openfl` should return:
`/projects/my/openfl/`

After we run `haxelib set openfl git`
`haxelib libpath openfl` should return:
`/usr/local/lib/haxe/lib/openfl/git/`

After we run `haxelib set openfl 3.6.1`
`haxelib libpath openfl` should return:
`/usr/local/lib/haxe/lib/openfl/3,6,1/`

Basically the `library manager` should inform us where `library` is currently located - which corresponds to the location where it was installed into.

back2dos:
I see. Well then I'd say it mostly looks good.

One thing to consider is that swallowing dependencies is maybe not a good idea (or there should be a way to opt out at least), because otherwise the caller may have to replicate that logic too.

mastef:
@back2dos I think I know what you mean - I would prefer to not touch any `path()` logic though, since it's so important. So I'd rather have `libpath` fail, than having `path` fail due to a logic change in `checkRec`. Although it's really just wrapping a conditional around [L1315-L1317](https://github.com/mastef/haxelib/blob/6e09f72a82c79b7da302ec567c9948d887cecfd5/src/haxelib/client/Main.hx#L1315-L1317) the return would still be a list with a `list.first()` call. So not much is gained I think.

mastef:
@back2dos dependencies are now not returned with the changes of c91c109

src/haxelib/client/Main.hx
mastef: @EricBishton the PR has now shifted to a separate method, so there's now a fully separate `libpath()` method, to not interfere with the `path` logic ( as both have different objectives ).

EricBishton:
Looks like it will do no harm to existing code paths.  I'd say it's good to go.

src/haxelib/client/Main.hx
kLabz: I meant there should not be a space before `:`, but as it's written this way all over the project (maybe because haxe has a french history :P), you can just ignore this comment :)

andyli:
For the version bump, there is no need to (read: don't) do so in PRs. Imagine there are multiple PRs bumping it, there will be conflicts. Also, since we're following semver, adding new feature means bumping "minor" instead of "bugfix". i.e. It should be 3.4.0.

Same for run.n, don't rebuild it in PRs since it will cause merge conflict.

Please add some example outputs in the documentation. It is not obvious what the output format will be when querying multiple libraries - paths can be separated by line-breaks, commas, colons (like $PATH) etc.

Finally, rebase the changes on top of current development to see if CI passes - I will merge if it does.

PR #413, opened 2018-03-14
Prompt to ignore proxy test failure (fixes #411)

### LINKED ISSUE #411: Allow forcing proxy configuration
I would like to be able to optionally configure the haxelib proxy setting without it pinging google.com. I had problems because my proxy configuration was correct, but the computer I was on didn't have access to google. Maybe it would be a good idea to have it say that it is trying to get to google when it does the test, too. I had to look at the code to find out why it thought it couldn't connect to the proxy.

RealyUniqueName:
see #411

### DIFF
--- a/src/haxelib/client/Main.hx
+++ b/src/haxelib/client/Main.hx
@@ -686,7 +686,7 @@ class Main {
 				doInstallFile(rep, prj, true, true);
 				return;
 			}
-			
+
 			if ( prj.endsWith("haxelib.json") )
 			{
 				installFromHaxelibJson( rep, prj);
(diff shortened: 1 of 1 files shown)


src/haxelib/client/Main.hx
```diff
@@ -1570,8 +1570,11 @@ class Main {
 		Http.PROXY = proxy;
 		print("Testing proxy...");
 		try Http.requestUrl("http://www.google.com") catch( e : Dynamic ) {
-			print("Proxy connection failed");
-			return;
+			var ignoreFail = param("Proxy connection failed. Use it anyway? (Y/n)");
```
andyli: Use `Cli.ask()` instead?

RealyUniqueName: Done

PR #442, opened 2019-03-07
Support haxe 4 `--library` and `-L` flags with haxelib install

haxiomic:
Running `haxelib install build.hxml` doesn't work if you specify your libraries with `-L` or `--library` (as suggested in the haxe 4 command doc)

This change matches `-lib`, `--library` and `-L` as library lines and fixes edge cases where invalid  flags like `-libx name` would pass but not correctly install

### DIFF
--- a/src/haxelib/client/Main.hx
+++ b/src/haxelib/client/Main.hx
@@ -712,9 +712,10 @@ class Main {
 							libsToInstall[lib] = { name: lib, version: null, type:"haxelib", url: null, branch: null, subDir: null }
 					}
 
-				if (l.startsWith("-lib"))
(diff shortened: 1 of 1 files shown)


Gama11:
I think this is only a half-solution, there can be arbitrary arguments in the `extraParams.hxml` of libraries.

haxiomic:
@Gama11 I'm not sure I see what you mean – can you give an example where there's a problem?

Gama11:
Nvm, I hadn't seen that this is `installFromHxml` (I thought this was another change for `haxelib path`). :)

haxiomic:
@Simn this one is worth merging for 4.0 because new users might be using the new `-L` and `--library` flags

PR #443, opened 2019-03-07
Fix bug where comments can interfere with `haxelib install`

haxiomic:
When running `haxelib install file.hxml`, comments in the hxml file can be matched as if they were arguments. For example

*file.hxml*:
```hxml
# this is a comment that ends in .hxml
```

Then running
`haxelib install file.hxml` 

Haxelib tries to load a file named "# this is a comment that ends in .hxml"

### DIFF
--- a/src/haxelib/client/Main.hx
+++ b/src/haxelib/client/Main.hx
@@ -701,15 +701,10 @@ class Main {
 		var libsToInstall = new Map<String, {name:String,version:String,type:String,url:String,branch:String,subDir:String}>();
 
 		function processHxml(path) {
-			var hxml = sys.io.File.getContent(path);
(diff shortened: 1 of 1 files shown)


Gama11:
Hm, I've needed to add some code to deal with comments here:

https://github.com/HaxeFoundation/haxelib/pull/439/files

Maybe there should be some helper method for this to avoid duplicating the logic?

haxiomic:
A helper method sounds good – maybe returning a list of flags + args. Ideally we want a fully specified hxml parser for this sort of thing but a helper method would be a start :)!

haxiomic:
I'll add something basic to this PR

haxiomic:
Nice :), to keep things simple for now I've just moved your strip and trip code into a separate method

PR #447, opened 2019-03-19
Fixes update and dependencies for libraries in VCS subdirectories 

jonasmalacofilho:
The main change is fixing the inconsistent use of .dev and .current that caused updates to fail when subdirectories are used.

The issue could be observed with mikestead's hx-yaml:

    > haxelib --never git yaml https://github.com/mikestead/hx-yaml master src
    Installing yaml from https://github.com/mikestead/hx-yaml branch: master
    Development directory set to /tmp/foobar/.haxelib/yaml/git/src

    > haxelib --never git yaml https://github.com/mikestead/hx-yaml master src
    You already have yaml version git installed.
    Updating yaml version git ...
    Error: [file_contents,/tmp/foobar/.haxelib/yaml/.current]

    Called from /home/jonas/Code/haxe/std/neko/_std/sys/io/File.hx line 30
    Called from haxelib/client/Main.hx line 1307
    Called from haxelib/client/Main.hx line 1476
    Called from a C function
    Called from haxelib/client/Main.hx line 1446
    Called from haxelib/client/Main.hx line 188
    Called from haxelib/client/Main.hx line 422

Additionally, dependencies were being silently ignored if subDir was not null, as the computation of the path to haxelib.json had been moved too far up.


jonasmalacofilho:
So many failures :open_mouth: 

...though none seem related these changes.

---

The legacy server seems to be broken on haxe development for a while.

hscript v2.3.0 causes the normal server to fail on Haxe v3.2.1 with:

    [...]/hscript/2,3,0/hscript/Parser.hx:758: characters 28-48 : Extractors in or patterns are not allowed

Haxe v3.1.3 started to fail in 0fe5e06 ("Get rid of macro-in-macro calls.") with:

    src/haxelib/SemVer.hx:157: characters 4-9 : null is not a valid version string

And on other cases the Mercurial tests most times fail with `CantCloneRepo`.  Maybe a network issue, or rate limiting?

jonasmalacofilho:
@Simn , can you please review/merge this fix?

(the failures on Travis are unrelated)

Simn:
I don't really know what's going on here but I trust you!

jonasmalacofilho:
Apparently, neither did I : (

Just noticed that this broke git library updates when `branch != null`.  I mistakenly though the existing tests already had that case covered...

@Simn, can you please revert the changes?  I rework the fix for `subDir != null` later.

PR #453, opened 2019-05-26
Better output format while installing. (MB)

eminfedar:
I know it is just a one time thing but...
Printing raw bytes number is not a good idea to give feedback about installing process to the user I think.

So I changed the output format like this. I hope you like it.
```
Old:
18739180/83729571 (22%)

New:
18.73 / 83.72 MB (22%)
```

Have a nice day :) 😄

### DIFF
--- a/src/haxelib/client/Main.hx
+++ b/src/haxelib/client/Main.hx
@@ -71,10 +71,12 @@ class ProgressOut extends haxe.io.Output {
 	function report(n) {
 		cur += n;
 		curMB = cur / 1000000.0;
+		curMB = Math.round( curMB * Math.pow(10, 2) ) / Math.pow(10, 2); // 12.34 precision.
 		if( max == null )
 			Sys.print(curMB + " MB\r");
 		else {
 			maxMB = max / 1000000.0;
+			maxMB = Math.round( maxMB * Math.pow(10, 2) ) / Math.pow(10, 2); // 12.34 precision.
 			Sys.print((curMB + " / " + maxMB + " MB (" + Std.int((curMB*100.0)/maxMB) + "%)\r");
 		}
 	}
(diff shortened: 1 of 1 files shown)


src/haxelib/client/Main.hx
```diff
@@ -68,10 +70,15 @@ class ProgressOut extends haxe.io.Output {
...
+		curMB = Math.round( curMB * Math.pow(10, 2) ) / Math.pow(10, 2); // 12.34 precision.
 		if( max == null )
-			Sys.print(cur+" bytes\r");
-		else
-			Sys.print(cur+"/"+max+" ("+Std.int((cur*100.0)/max)+"%)\r");
+			Sys.print(curMB + " MB\r");
+		else {
+			maxMB = max / 1000000.0;
+			maxMB = Math.round( maxMB * Math.pow(10, 2) ) / Math.pow(10, 2); // 12.34 precision.
+			Sys.print((curMB + " / " + maxMB + " MB (" + Std.int((curMB*100.0)/maxMB) + "%)\r");
```
Aurel300: Maybe keep the `%` calculated from `cur` and `max`?

src/haxelib/client/Main.hx
```diff
@@ -68,10 +70,15 @@ class ProgressOut extends haxe.io.Output {
 
 	function report(n) {
 		cur += n;
+		curMB = cur / 1000000.0;
+		curMB = Math.round( curMB * Math.pow(10, 2) ) / Math.pow(10, 2); // 12.34 precision.
```
Aurel300: Why not `100`? I don't think it's more readable this way, and it's marginally slower.

eminfedar:
Changes are coming.

Aurel300:
The changes you are introducing are pretty small, so they could be a part of a single PR, but then you should edit the title. Additionally, you should compile the code locally before pushing anything – anything you push to a PR notifies us (another reason to keep PRs atomic). The latest commit also has syntactic and typing errors.

eminfedar:
Yeah you are right. I will open a new PR. We can close this :)

Gama11:
>better to use mebibytes (MiB) rather than megabytes?

Why?

PR #454, opened 2019-05-27
feature: Better 'install' command output.

eminfedar:
I have tested this one and works well. Also considered the advices. :)

### DIFF
--- a/src/haxelib/client/Main.hx
+++ b/src/haxelib/client/Main.hx
@@ -71,15 +71,15 @@ class ProgressOut extends haxe.io.Output {
 	function report(n) {
 		cur += n;
 		
-		var tag : String = ((max != null ? max : cur) / (1024*1024)) > 1 ? "MiB" : "KiB";
+		var tag : String = ((max != null ? max : cur) / 1000000) > 1 ? "MB" : "KB";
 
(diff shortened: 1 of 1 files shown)


Gama11:
Is MiB / KiB really better than MB / KB though? I think most people don't really have much of a concept of the former units...

back2dos:
Nitpick: it's [`kB`](https://en.wikipedia.org/wiki/Kilo-).

RealyUniqueName:
I'm too old for MiB/KiB and prefer classic MB/KB naming too.
But aparently MB doesn't mean 2^10 anymore, because HDD manufacturers and developers of desktop environments decided that MB is 10^6.

eminfedar:
So, what should I use, 2^n or 10^n calculation with "MB/KB" tags?

haxiomic:
I think 10^n is the way to go for MB/KB (see wikipedia page https://en.wikipedia.org/wiki/Megabyte)

Aurel300:
@eminfedar I think I'm outvoted here! Sorry for wasting your time on this, but I guess it should read `MB` for multiples of `1000000`, and `kB` (lowercase `k` is more correct) for multiples of `1000`.

Issue #451, opened 2019-05-07
Haxelib should error if there's too many arguments for a command


Gama11:
At least `haxelib version` accepts (and ignores) additional arguments, which can be extremely misleading:

```haxe
λ haxelib version
3.4.0

λ haxelib version hxcpp
3.4.0
```

PR #455, opened 2019-05-27
fix: #451

eminfedar:


Gama11:
Hm, a few thoughts:

- shouldn't the warning be printed to stderr instead of stdout?
- I'm not sure why the version would still be printed in case of too many parameters?
- I think the error message could be clearer, something like "no parameters expected, got <parameters>".

src/haxelib/client/Main.hx
```diff
@@ -244,7 +244,10 @@ class Main {
 	}
 
 	function version() {
-		print(VERSION_LONG);
+		Sys.println('Haxelib\'s version: ${VERSION_LONG}');
+
+		if ( paramOpt() != null )
+			Sys.println('! WARNING: "version" parameter only shows the version of the "haxelib"');
```
RealyUniqueName: That's a misleading error.
What if a user typed `haxelib version --verbose`?
The message should tell exactly, what's the problem. And the problem is invalid amound of arguments.

src/haxelib/client/Main.hx
```diff
@@ -244,7 +244,10 @@ class Main {
 	}
 
 	function version() {
-		print(VERSION_LONG);
+		Sys.println('Haxelib\'s version: ${VERSION_LONG}');
```
RealyUniqueName: It's better not to change output format for no reason. Some tools or CIs may rely on it.

src/haxelib/client/Main.hx
eminfedar: I can't get the difference with tag `--verbose`.

I think the only usage is `haxelib version`, isn't it?

RealyUniqueName: It's just an example of a case when proposed error message would be misleading.

eminfedar:
This is okay I think? -> feb175f

RealyUniqueName:
That's better, but needs a newline in the end of the message and `Sys.exit(1)` after the message :)

eminfedar:
Should `print(VERSION_LONG)` line exit with 0 too?

RealyUniqueName:
No, that happens by default.

PR #456, opened 2019-06-02
Add --skip-dependencies option

### LINKED ISSUE #343: --skip-dependencies option in haxelib
When installing an extension with a dependency like `"openfl": ""`, haxelib starts downloading the latest version ( e.g. `4.0.3` ) although openfl is already present on the system ( `3.6.1` )

Now there would be two/three ways to go about this : 
- when no version number is specified, and the library exists already, use the existing library. But the drawback is that it's usually actually nice to have the latest libraries installed. So I'd rather advocate for keeping this behavior
- allow an option like `--skip-dependencies` that lets the user install dependencies on their own if they wish to
- allow an option like `--no-update` which only checks for dependencies being met, and doesn't auto-update to latest, unless dependency demands it ( `>=4` )

Having written the last paragraph actually just noticed that dependencies can't specify a version number equal to, or above. That would be a good enhancement too

kLabz:
Fixes #343 with the most basic variant: just don't install any dependency listed in haxelib.json when this flag is set.

**Background:** I needed something like this to be able to use some libs meant to be used with lix and which cannot be installed without lix due to dependencies (conflicting versions or even libs never released on haxelib) [[Gist with scripts used for that]](https://gist.github.com/kLabz/e29722b3425ad2f36341d86d3612c117)

Updated help output:
```
Haxe Library Manager 3.4.0 - (c)2006-2019 Haxe Foundation
  Usage: haxelib [command] [options]
  Basic
    install            : install a given library, or all libraries from a hxml file
    update             : update a single library (if given) or all installed libraries
    remove             : remove a given library/version
    list               : list all installed libraries
    set                : set the current version for a library
  Information
    search             : list libraries matching a word
    info               : list information on a given library
    user               : list information on a given user
    config             : print the repository path
    path               : give paths to libraries' sources and necessary build definitions
    libpath            : returns the root path of a library
    version            : print the currently used haxelib version
    help               : display this list of options
  Development
    submit             : submit or update a library package
    register           : register a new user
    dev                : set the development directory for a given library
    git                : use Git repository as library
    hg                 : use Mercurial (hg) repository as library
  Miscellaneous
    setup              : set the haxelib repository path
    newrepo            : create a new local repository
    deleterepo         : delete the local repository
    convertxml         : convert haxelib.xml file to haxelib.json
    run                : run the specified library with parameters
    proxy              : setup the Http proxy
  Available switches
    --flat             : do not use --recursive cloning for git
    --always           : answer all questions with yes
    --debug            : run in debug mode, imply not --quiet
    --quiet            : print less messages, imply not --debug
    --system           : run bundled haxelib version instead of latest update
    --skip-dependencies: do not install dependencies
    --never            : answer all questions with no
    --global           : force global repo if a local one exists
```

### DIFF
--- a/src/haxelib/client/Main.hx
+++ b/src/haxelib/client/Main.hx
@@ -250,7 +250,8 @@ class Main {
 	function usage() {
 		var cats = [];
 		var maxLength = Lambda.fold(Reflect.fields(ABOUT_SETTINGS), function(opt, max) {
-			var len = opt.length + 2; // include -- in option name
+			var fullOption = '--' + ~/([A-Z])/g.replace(opt, "-$1").toLowerCase();
+			var len = fullOption.length;
 			return len > max ? len : max;
 		}, 0);
 
@@ -273,8 +274,10 @@ class Main {
 		}
 
 		print("  Available switches");
-		for (f in Reflect.fields(ABOUT_SETTINGS))
-			print('    --' + f.rpad(' ', maxLength-2) + ": " + Reflect.field(ABOUT_SETTINGS, f));
+		for (f in Reflect.fields(ABOUT_SETTINGS)) {
+			var option = ~/([A-Z])/g.replace(f, "-$1").toLowerCase().rpad(' ', maxLength-2);
+			print('    --' + option + ": " + Reflect.field(ABOUT_SETTINGS, f));
+		}
 	}
 	static var ABOUT_SETTINGS = {
 		global : "force global repo if a local one exists",
--- a/src/haxelib/client/Main.hx
+++ b/src/haxelib/client/Main.hx
@@ -249,7 +249,11 @@ class Main {
 
 	function usage() {
 		var cats = [];
-		var maxLength = 0;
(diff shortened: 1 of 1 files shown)


RealyUniqueName:
> I went for --skipdeps instead of --skip-dependencies not to break the current help message...

I'm not sure that's a valid reason for weird option naming :)
I'ts better to make help message automatically adjustng to the longest option name.

kLabz:
Yeah.. I just didn't want to have to do this x)

kLabz:
Ok the helper was already there actually, it was just ignoring switches.

kLabz:
Damn, need to convert lowerCase to snake_case in `usage()`...

kLabz:
@RealyUniqueName done :)

Gama11:
Hm, it seems existing commands don't separate words at all: `convertxml`, `newrepo`...

Aurel300:
@Gama11 Those are commands though. I am in favour of having `--skip-dependencies` with the hyphen to be consistent with `--` switches in general.

PR #459, opened 2019-06-27
Using more current compiler options

uvtc:
`-L|--library` rather than `-lib`. Similarly for `-js` to `--js`.


RealyUniqueName:
Thanks

PR #462, opened 2019-08-15
Adding a timeout to the socket connection

qwertykg:
**The Problem**
We are currently experiencing major problems when trying to run "haxelib install haxelib.json". When running this command. At a certain **random** point during the install the command just hangs forever. This happens intermittently and we think that this could either be caused by stress on the network we use/ latency from being too far away from the server or an internet routing issue. 

As this happens randomly and not in any consistent manner, we noticed that most of the time, you can trigger a retry and the second time round everything is perfectly fine. Thus implementing some kind of retry feature would probably work quite well for us and possibly others. 

We noticed that the stall always happens in one of three places:

1. `On line 916`: `h.customRequest(false, progress);`. More specifically, in the customRequest function when `sock.connect(new Host(host),port);` is called. This function gets called and just hangs forever. Since there is no timeout on the socket, no error gets thrown and nothing triggers a retry

2. `On the line 957`: `site.postInstall(project, version);`. From what I can see, this function simply makes a request to the database to receive the information about the `project` and then increments the number of times this `project` was downloaded and updates the database. Unfortunately this is server side code, so I don't know how I can debug this file to determine the exact line that stalls. I'm assuming it's whatever bit of code makes the actual database connection that stalls.

3. `On line 1053`: `d.version = site.getLatestVersion(d.name);`. This function just does a select sql query to get the latest version. I'm also assuming the stall happens when a connection to the database is established.

**The Proposed Solution**
As a solution, i made a change to fix the first case. Instead of letting customRequest create a new socket, I am creating it inside Main.hx so that I can set a timeout on this socket and pass the socket through as the third parameter. And so if the timeout is reached before the library is downloaded an error is thrown which triggers a retry to download the library. From what we can see on our side the library always downloads successfully on the second try. With regards too the timeout, I've arbitrarily chosen 10 seconds but I'm happy to increase that time to whatever you guys think would be acceptable. I do understand that some users might have slow internet and their libraries could be taking more than 10 seconds to download without getting stuck.

I do however require assistance with cases 2 and 3. From what I understand, i dont think I can modify this file and see my changes locally? So I wasn't able to implement some sort of timeout feature in this file. Would you guys also be able to implement this as part of this pull request? Or give me some direction as to how i can modify this file and see my changes locally so that I can try and get something working?


Simn:
The approach here doesn't work at the moment due to https://github.com/HaxeFoundation/haxe/issues/7619. We'll get back to you with further information shortly.

Simn:
Something you could try is moving [`sock.setTimeout(cnxTimeout);`](https://github.com/HaxeFoundation/haxe/blob/3.4.7/std/haxe/Http.hx#L590) inside [this branch](https://github.com/HaxeFoundation/haxe/blob/3.4.7/std/haxe/Http.hx#L433). What maybe happens here is that the timeout is set too late. Plus it seems correct to set the socket timeout if the socket instance is created within Http.

qwertykg:
Sure :) I will do this and submit the PR on the haxe repo - Are you happy with the cnxTimeout being 10 seconds? 

Also do you have any idea's on how to tackle issues 2 and 3 in the description? This code doesn't seem easily debuggable as it resides server side, so I can't reliably suggest a change that would trigger a retry if those parts hang.

Simn:
It's hard to tell, but 2 and 3 are probably related to this as well. I think we should set the socket timeout and then see if the problem persists.

By the way, what Haxe version do you guys use?

qwertykg:
Cool, will do. So we're all currently on version 4 rc-3. We try to stay updated as much as possible.

qwertykg:
Also, the branch you pointed me to is actually a tag, do I branch off that tag and make the change? In which case which branch do I merge it back into in the Haxe repo? Or do I just use that branch to build the new run.n and merge the new run.n into the Haxelib repo?

Simn:
Ah I just linked the Haxe 3.4.7 branch because that's usually what people use in production. Since you're on Haxe 4 already, we can just make the change from the `development` branch. In that case the relevant file is here: https://github.com/HaxeFoundation/haxe/blob/development/std/sys/Http.hx

qwertykg:
Cool, thank you :) I have submitted the PR here: https://github.com/HaxeFoundation/haxe/pull/8646.

ncannasse:
I this PR still up to date now that this has been handled at the Http level?

Simn:
No we can close here, thanks for reminding me.

PR #463, opened 2019-08-19
Add retry functionality to SiteApi

qwertykg:
After adding a default timeout to the HTTP class in this PR: https://github.com/HaxeFoundation/haxe/pull/8646, we noticed that case number 2 and 3 described in this PR: https://github.com/HaxeFoundation/haxelib/pull/462 was still an issue as when the socket times out, it fails the command instead of triggering a retry. I have added changes to implement a retry and tested this locally and it seems to work fine on our repo's. 

I do understand this is a big change to merge in, as every call to the SiteApi class is now going to infinitely try and make a connection. And if the connection never succeeds it could get stuck. If you guys have a different suggestion for this, I will happily reimplement the change - Or perhaps hide this change behind a flag? Either way, I would need some direction as to what you guys think will be cool to get merged in :)

Also - the pipeline seems to be failing with some dirty source error - How do I fix this? I built run.n using the latest nightly build.


Simn:
Some remarks:

* A retry is reasonable, but doing it infinitely is not. Typically you would have a `var numTries = 3;` and then loop with `while (numTries-->0)`. 
* With the `catch(e:Dynamic)` we run the risk of retrying for the wrong reasons. That is, if the called function throws an exception that has nothing to do with connection timeouts, we would keep retrying. @Aurel300 Do you know if we can catch something more specific here?
* Having all these `WithRetry` functions is a bit of an unnecessary abstraction. Consider a pattern like this:

```haxe
class Main {
	static public function main() {
		retry(failRandomly.bind("myArgument"), 3);
	}

	static function failRandomly(argument:String) {
		trace('Trying failRandomly($argument)');
		if (Math.random() > 0.3) {
			throw "fail";
		}
		return 'result: $argument';
	}

	static function retry<R>(f:() -> R, numTries:Int = 1) {
		while (numTries-->0) {
			try {
				return f();
			} catch(e:Dynamic) {
				trace('Failed because $e');
			}
		}
		throw 'Failed after multiple tries';
	}
}
```

Simn:
Also I'm checking with the team what that dirty source is about (cc @andyli).

qwertykg:
Thanks :) I have implemented the changes - Will make more changes to fix the CI and to catch something more specific when we get the feedback on those. In the mean time, let me know if there's more changes i should make :)

Aurel300:
Please delete `run.n` from the commit. I'm not sure why you have commits that delete it immediately followed by re-uploading it. I believe that it is what is causing the CI failure.

qwertykg:
Ahh sorry - I have deleted it now, do i need to commit the original run.n back? Or does it get automatically created?

qwertykg:
So I figured out what dirty source means - The "dirty" string is added to the haxelib version if you build run.n with uncommitted changes 😛 I have commited everything then built my run.n - It should be fine now

Simn:
Right, I somehow thought this was a git error... CI looks good now.

I still want to check the `catch(Dynamic)` before I merge.

Also, I suppose a retry function which defaults to 1 try is a bit silly. We can change the default value to 3 and remove the 3 from all the calls to make it look a bit nicer.

qwertykg:
Cool I've made the update to change the default retries to 3 and removed the second parameter from the function calls :)

Simn:
Nice, the diff looks good to me now!

Our CI seems to be having some networking problems at the moment which is somewhat ironic.

src/haxelib/client/Main.hx
```diff
@@ -221,8 +221,19 @@ class Main {
 		initSite();
 	}
 
+	function retry<R>(func:Void -> R, numTries:Int = 3) {
+		while (numTries-- > 0) {
+			try {
+				return func();
+			} catch(e:Dynamic) {
+				print('Failed with error: $e');
```
RealyUniqueName: I think it's worth to print that we're about to retry the operation and print a success message in case of successful retry.
If user would see `Failed with error` and no success message afterwards, user would think it failed.

Simn:
Assigning for the catch-question.

qwertykg:
I have added better logging as per @RealyUniqueName's request. Could you verify that this new change is acceptable? I also added an if statement to only allow for retries if a HTTP timeout occured. I think that should cover the problem the "catch-question" was trying to solve? Let me know if there's anything else I should change :)

RealyUniqueName:
LGTM

qwertykg:
Cool - thank you. I've resolved your discussion. Does anyone know why the pipeline seems to be failing?

andyli:
The build passed after I restarted it. So it was most likely network problem.

qwertykg:
Ah thats great :) @Simn is there any other change you would like to add here?

Simn:
We're good, thanks for looking into this! Let me know if there's anything else we can do to help.

I'm also still interested in why you guys are having these networking problems in the first place and if there's something we can do about that long-term (e.g. different hosting).

PR #467, opened 2019-09-17
Support lib:dev:/path in installFromHxml

kLabz:
`haxelib install install.hxml` currently supports things like `-lib something:git:git-url#reference`

This PR adds support for `-lib something:dev:../libs/something` too (mapping to `haxelib dev something ../libs/something`)


src/haxelib/client/Main.hx
```diff
@@ -48,6 +48,16 @@ private enum CommandCategory {
 	Deprecated(msg:String);
 }
 
+private typedef LibInstallationInfo = {
+	name:String,
+	version:String,
+	type:String,
```
Aurel300: At this point it would probably be best to change `type` into an enum, with cases `Dev(path:String)`, `Git(url:String, branch:String, subDir:String)`, `Haxelib`.

kLabz: Yeah there's many things that can be rewritten.. But are we doing it?

Aurel300: 🤷‍♂ yeah there are. Rewriting this part in particular in this PR seems like a convenient opportunity though.

kLabz: I knew I couldn't stop if I started.. Damn I don't know where this will end x)

PR #470, opened 2019-10-04
Drop support for Haxe < 3.4

RealyUniqueName:


andyli:
Please no. The haxelib server has to be tested with older versions of the client or else one day the eariler Haxe installation will not be able to use lib.haxe.org.

Issue #498, opened 2020-05-13
'haxelib submit' does not work as documented


notboring:
When submitting a local directory to haxelib using 'haxelib submit' I was asked to enter a package name. When entering the name of the package I received a file error. 
I was a bit confused at first but after trying a few things I noticed that you need to pass a path along with the submit command if you want to commit the local directory:
'haxelib submit .' 
The documentation states 'Alternatively you can run haxelib submit without argument to have Haxelib zip and submit the current directory (excluding names starting with a dot).' so I wonder if this is an issue with the docs or with the current haxelib version. (or maybe with my local installation?)

This is on windows 10 using Haxe 4.0.5.

RealyUniqueName:
I can't find the code in haxelib client which would do that. It's probably a misleading documentation.

RealyUniqueName:
But if the argument is a path to an existing directory, then haxelib client should indeed zip and submit it:
https://github.com/HaxeFoundation/haxelib/blob/e7cb62181303eb61737fe06b5784cab41a58a53f/src/haxelib/client/Main.hx#L625-L629

notboring:
Thanks for looking into this. I submitted a small change to the haxelib documentaiton.

PR #499, opened 2020-05-13
'haxelib submit' needs an additional parameter

notboring:
Zipping + submitting local directories as described in the docs won't work without providing the path of the directory.
Original issue: https://github.com/HaxeFoundation/haxelib/issues/498


RealyUniqueName:
Thanks!

PR #523, opened 2021-09-17
Avoid using system shell to start processes

justin-espedal:
This relates to https://github.com/HaxeFoundation/haxe/issues/10389

This change allows Haxe's basic build tooling to be used even when the system shell is unavailable.

`Sys.command` runs processes through the system shell. The 2-argument form of `new Process`, on the other hand, starts processes without going through the shell.

### DIFF
--- a/src/haxelib/client/Main.hx
+++ b/src/haxelib/client/Main.hx
@@ -1660,6 +1660,37 @@ class Main {
 	}
 	static var __haxeVersion:SemVer;
 
+	function createProcess(command:String, args:Array<String>)
+	{
+		var process = new Process(command, args);
+		
+		var waiting = true;
+
+		while (waiting)
+		{
+			try
+			{
+				Sys.println(process.stdout.readLine());
+			}
+			catch (e:haxe.io.Eof)
+			{
+				waiting = false;
+			}
+		}
+
+		var error = process.stderr.readAll().toString();
+		var result = process.exitCode();
+
+		if (error != "")
+		{
+			Sys.println(error);
+		}
+		
+		process.close();
+
+		return result;
+	}
+
 	function doRun( rep:String, project:String, version:String ) {
 		var pdir = rep + Data.safe(project);
 		if( !FileSystem.exists(pdir) )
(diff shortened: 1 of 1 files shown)


RealyUniqueName:
This is not equal to `Sys.command` as it prints stdout data as it arrives, but accumulates stderr messages until the process is finished. Moreover stdin is completely ignored.

back2dos:
There are rather difficult problems to overcome here:

1. On windows, haxe and neko can be `.cmd` files when installed through npm (via [haxe](https://www.npmjs.com/package/haxe) or [lix](https://www.npmjs.com/package/lix). Or they may be `.bat` files if the user has their own stuff set up for whatever reason. The resolution of executables other than `.exe` relies on the shell, meaning you'll have to resolve the right one via `where` (assuming it's even available to you - otherwise you'll have to locate it in cwd + `PATH` by hand).

2. Dealing with stdio is important, because some haxelibs have *interactive*, e.g. [hxcpp](https://github.com/HaxeFoundation/hxcpp/blob/e23515d4acd43f3609e5ca49a8cebdad285db78d/tools/run/RunMain.hx#L46). A few libraries even call haxelib, e.g. a [submission helper I created](https://lib.haxe.org/p/submit) which ultimately calls haxelib and runs into `Sys.getChar(false)` which requires the haxelib process's stdin to be connected to a terminal - `Sys.command` accomplishes this if the current process's stdin is also connected to a terminal, because child processes created in this fashion inherit std. The current process API doesn't allow for it, but [future APIs will](https://github.com/HaxeFoundation/haxe/pull/9111/files#diff-8c0f9abcc580130aca774e6932ea42a3c108d0c1dc59bd6d1e27bf6c5e06dfb1). 
   You *could* attempt to run haxelib on a platform where an API is available to inherit stdio to child processes, e.g. via libuv on hl or eval (although this means there'll be no binary per se, only a `.bat` file calling the haxe interpreter) or java (using a [native image](https://www.graalvm.org/reference-manual/native-image/) or [a packaged executable with a JRE](https://github.com/dgiagio/warp)) or cs (not too familiar with .NET but I expect it has such capabilities).

PR #525, opened 2021-10-01
Add "why" command

TheDrawingCoder-Gamer:
Adds "why" command, which finds the reason a library must be included. Basic documentation:
`haxelib why [library]`
Looks through the haxelib.json dependecies, and returns the dependency path if it's found, or states it couldn't find the library.
`haxelib why [library] [hxml]` 
Looks through the hxml file and finds why a library is required, returning the dependency path if it's found, or states it couldn't find the library. Follows `--library` tags and hxml references.

Similar to yarn's [why command](https://yarnpkg.com/cli/why)

Optimization: This probably needs optimization, as it traverses the tree of dependencies of a tree. Threading would be helpful

Further Improvements: Right now, it only returns 1 path, when it may be required in multiple places.


TheDrawingCoder-Gamer:
Oh, it can be uppercased??

RealyUniqueName:
I don't see a button to run CI for PRs. @Simn , do you have it?

Simn:
No

TheDrawingCoder-Gamer:
Will you guys handle the run.n? it seems like a recompile will be required anyway

RealyUniqueName:
You need to merge development and avoid committing `run.n` after that.

PR #540, opened 2022-01-07
Fix tests

tobil4sk:
Clean up integration tests' output, and add test configuration for vscode.


test/TestBase.hx
```diff
@@ -1,16 +1,18 @@
...
+	static final haxelibPath = FileSystem.fullPath("run.n");
+
 	public function runHaxelib(args:Array<String>, printProgress = false) {
-		var p = new Process("neko", [haxelibPath].concat(args));
+		final p = new Process("neko", [haxelibPath].concat(args));
 		var stdout = '';
 		var stderr = '';
 		var eofCount = 0;
-		var c;
+		var c:Int;
```
Simn, line 15: Was this necessary for some reason?

test/tests/TestSemVer.hx
```diff
@@ -61,12 +62,9 @@ class TestSemVer extends TestBase {
...
 
-	function parseInvalid( str:String ) {
-		return try {
-			(SemVer.ofString( str ) : String);
-		} catch (e:String) {
-			"invalid";
-		}
+	function parseInvalid( str:String ):String {
+		return try SemVer.ofString(str)
+			catch (e:String) "invalid";
```
Simn, line 67: Indentation looks weird here.

test/HaxelibTests.hx
```diff
@@ -20,38 +19,37 @@ class HaxelibTests {
...
 		if (!FileSystem.exists(dir)) return;
-		var exitCode = switch (Sys.systemName()) {
+		final exitCode = switch (Sys.systemName()) {
 			case "Windows":
 				Sys.command("rmdir", ["/S", "/Q", StringTools.replace(FileSystem.fullPath(dir), "/", "\\")]);
 			case _:
 				Sys.command("rm", ["-rf", dir]);
 		}
-		if (exitCode != 0) {
+		if (exitCode != 0)
```
Simn, line 40: Personally, I don't like this brace-less style. But if this what the rest of haxelib does then I agree that we should make it consistent.

test/tests/TestSemVer.hx
tobil4sk: Reverted

test/HaxelibTests.hx
tobil4sk: I had a look through and it does seem like the braces are preferred. Might be a good idea to setup a haxe checkstyle file for stuff like this at some point.

test/TestBase.hx
tobil4sk: No, reverted it now.

PR #547, opened 2022-04-03
Haxe 4 updates

tobil4sk:
Get rid of Std.is warning and use final instead of var

### DIFF
--- a/src/haxelib/client/ConvertXml.hx
+++ b/src/haxelib/client/ConvertXml.hx
@@ -26,7 +26,7 @@ import haxe.Json;
 class ConvertXml {
 	public static function convert(inXml:String) {
 		// Set up the default JSON structure
-		var json = {
+		final json = {
 			"name": "",
 			"url" : "",
 			"license": "",
@@ -39,8 +39,8 @@ class ConvertXml {
 		};
 
 		// Parse the XML and set the JSON
-		var xml = Xml.parse(inXml);
-		var project = xml.firstChild();
+		final xml = Xml.parse(inXml);
+		final project = xml.firstChild();
 		json.name = project.get("name");
 		json.license = project.get("license");
 		json.url = project.get("url");
@@ -58,9 +58,14 @@ class ConvertXml {
 						case "description":
 							json.description = node.firstChild().toString();
 						case "depends":
-							var name = node.get("name");
-							var version = node.get("version");
-							if (version == null) version = "";
+							final name = node.get("name");
+							final version = {
+								final version = node.get("version");
+								if (version == null)
+									"";
+								else
+									version;
+							}
 							Reflect.setField(json.dependencies, name, version);
 						default:
 					}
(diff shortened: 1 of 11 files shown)


src/haxelib/client/ConvertXml.hx
```diff
@@ -58,9 +58,14 @@ class ConvertXml {
 						case "description":
 							json.description = node.firstChild().toString();
 						case "depends":
-							var name = node.get("name");
-							var version = node.get("version");
-							if (version == null) version = "";
+							final name = node.get("name");
+							final version = {
```
Simn, line 62: "With a final in hand, everything looks like a countdown"

Frankly, for cases like this I prefer sticking to `var` instead of making awkward workarounds just because...

src/haxelib/client/Main.hx
```diff
@@ -178,15 +180,16 @@ class Main {
...
+	final alreadyUpdatedVcsDependencies:Map<String,String> = new Map<String,String>();
 
-	var argcur : Int;
-	var args : Array<String>;
-	var commands : List<{ name : String, doc : String, f : Void -> Void, net : Bool, cat : CommandCategory }>;
-	var siteUrl : String;
-	var site : SiteProxy;
-	var isHaxelibRun : Bool;
-	var alreadyUpdatedVcsDependencies:Map<String,String> = new Map<String,String>();
+	var argcur:Int;
```
Simn, line 189: The only diff here is the loss of spaces, but we're not being consistent in that regard (e.g. `final commands` just above).

src/haxelib/client/Main.hx
```diff
@@ -1283,36 +1289,40 @@ class Main {
...
 
 	function list() {
-		var rep = getRepository();
-		var folders = FileSystem.readDirectory(rep);
-		var filter = paramOpt();
-		if ( filter != null )
-			folders = folders.filter( function (f) return f.toLowerCase().indexOf(filter.toLowerCase()) > -1 );
-		var all = [];
+		final rep = getRepository();
+		final folders = {
```
Simn, line 1303: Just to mention a counterpoint, this one I actually like because it hides `filter` in a sub-scope, which is good because we don't need it afterwards. I'll have to get used to these nested `final something = { final something = }`, but I don't think it's a bad pattern per-se (I use it in OCaml too).

src/haxelib/client/ConvertXml.hx
tobil4sk: Hm, is it fine to test with haxe nightly so we can use the null coalescing operator here?

src/haxelib/client/Main.hx
```diff
@@ -1586,35 +1600,47 @@ class Main {
 		}
 	}
 
-
 	function removeExistingDevLib(proj:String):Void {
 		//TODO: ask if existing repo have changes.
 
 		// find existing repo:
-		var vcs = Vcs.getVcsForDevLib(proj, settings);
+		var vcs = Vcs.getVcsForDevLib(proj, {
```
Simn, line 1607: Why are we creating a new object here?

src/haxelib/client/Main.hx
tobil4sk: Reverted for the unchanged members

PR #554, opened 2022-04-11
Fix haxelib.json issues

### LINKED ISSUE #488: No option for non-open source license
It is not possible to use haxelib for internal libraries without giving your code an open-source license.
At least one option should be for internal code, or it should simply not check the license at run time. (is it really needed when doing a 'haxelib path lib' ?) ie. just check on the server when uploading.
Possible extra licenses types:  private, commercial, internal, custom

tobil4sk:
- Do not impose MIT when haxelib.json parsing fails.
- Add tests for #488
- Do not allow git dependencies when submitting a library. See: https://github.com/HaxeFoundation/haxelib/pull/344#issuecomment-244006799

Closes #488

### DIFF
--- a/src/haxelib/Util.hx
+++ b/src/haxelib/Util.hx
@@ -57,12 +57,4 @@ class Util {
 			return macro $v{version};
 		}
 	}
-
-	public static macro function replaceIfNull(toReplace:Expr, with:Expr)
-		return macro if ($toReplace == null)
-			$toReplace = $with;
-
-	public static macro function replaceIf(toReplace:Expr, with:Expr, condition:Expr)
-		return macro if ($condition)
-			$toReplace = $with;
 }
--- a/src/haxelib/Util.hx
+++ b/src/haxelib/Util.hx
@@ -57,4 +57,12 @@ class Util {
 			return macro $v{version};
 		}
 	}
+
+	public static macro function replaceIfNull(toReplace:Expr, with:Expr)
+		return macro if ($toReplace == null)
+			$toReplace = $with;
+
+	public static macro function replaceIf(toReplace:Expr, with:Expr, condition:Expr)
+		return macro if ($condition)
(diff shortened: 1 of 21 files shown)


src/haxelib/Util.hx
```diff
@@ -28,33 +28,41 @@ class Util {
...
 		try {
-			//check if the .git folder exist
-			//prevent getting the git info of a parent directory
-			if (!sys.FileSystem.isDirectory(".git"))
-				throw "Not a git repo.";
-
 			//get commit sha
 			p = new sys.io.Process("git", ["rev-parse", "HEAD"]);
-			final sha = p.stdout.readAll().toString().trim();
+			var sha = p.stdout.readAll().toString().trim();
```
Simn, line 40: Huh?

src/haxelib/Util.hx
```diff
@@ -28,33 +28,41 @@ class Util {
...
 			}
-			return macro $v{version};
+			return macro $v{longVersion};
 		} catch(e:Dynamic) {
 			if (p != null) p.close();
 			return macro $v{version};
 		}
 	}
+
+	public static macro function replaceIfNull(toReplace:Expr, with:Expr)
```
Simn, line 61: Let's please not do this. Duplicating expressions is not sound in the general case, so this is a potential trap.

src/haxelib/Util.hx
tobil4sk: That file got moved into the top level `haxelib` package, which causes it to get picked up when the server is compiling (which requires haxe 3).

I'm not sure why it wasn't failing before, but now it did:
https://github.com/HaxeFoundation/haxelib/runs/5971785204?check_suite_focus=true

src/haxelib/Util.hx
tobil4sk: Reverted.

PR #555, opened 2022-04-11
Print which haxelib is used when in debug mode

tobil4sk:
Closes #345


Simn:
Does this affect the `haxelib path` calls from the compiler?

tobil4sk:
Technically no, because the compiler calls don't pass in `--debug`. However, perhaps there is an argument to be had about whether this belongs in stdout or stderr.

tobil4sk:
> Does this affect the haxelib path calls from the compiler?

> Technically no, because the compiler calls don't pass in --debug

Looks like `lime` does pass in `--debug` to haxelib during debug builds, which causes this to break builds there... Might be better to print this to stderr instead.

PR #557, opened 2022-04-14
Add support for `hg` dependencies

tobil4sk:
Merge code which parses `git:https://some.url` and `somelib:git:https://some.url`, thus allowing for mercurial dependencies in `haxelib.json` files (this was missing for some reason). Includes appropriate testing.

### DIFF
--- a/src/haxelib/VersionData.hx
+++ b/src/haxelib/VersionData.hx
@@ -56,19 +56,17 @@ class VersionDataHelper {
 			return Haxelib(SemVer.ofString(versionInfo));
 		} catch (_:String) {}
 
-		try {
-			var data = getVcsData(versionInfo);
-			return VcsInstall(data.type, data.data);
-		} catch (e:String) {
+		var data = getVcsData(versionInfo);
+		if (data == null)
 			throw '$versionInfo is not a valid library version';
-		}
+		return VcsInstall(data.type, data.data);
 	}
 
 	static var vcsRegex = ~/^(git|hg)(?::(.+?)(?:#(?:([a-f0-9]{7,40})|(.+)))?)?$/;
 
-	static function getVcsData(s:String):{type:VcsID, data:VcsData} {
+	static function getVcsData(s:String):Null<{type:VcsID, data:VcsData}> {
 		if (!vcsRegex.match(s))
-			throw '$s is not valid';
+			return null;
 		var type = switch (vcsRegex.matched(1)) {
 			case Git:
 				Git;
--- a/src/haxelib/VersionData.hx
+++ b/src/haxelib/VersionData.hx
@@ -2,8 +2,8 @@ package haxelib;
 
 /** Abstract enum representing the types of Vcs systems that are supported. **/
 @:enum abstract VcsID(String) to String {
-	final Hg = "hg";
-	final Git = "git";
+	var Hg = "hg";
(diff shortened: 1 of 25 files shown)


src/haxelib/api/Connection.hx
```diff
@@ -417,12 +417,12 @@ class Connection {
 	}
 
 	static function checkDependencies(dependencies:Dependencies) {
-		for (d in dependencies) {
-			final versions:Array<String> = getVersions(ProjectName.ofString(d.name));
-			if (d.version == "")
+		for (name => versionString in dependencies) {
```
Simn, line 420: Wasn't there something that needed to be compilable with Haxe 3, which doesn't have this syntax?

src/haxelib/VersionData.hx
```diff
@@ -0,0 +1,89 @@
...
+class VersionDataHelper {
+	public static function extractVersion(versionInfo:String):VersionData {
+		try {
+			return Haxelib(SemVer.ofString(versionInfo));
+		} catch (_:String) {}
+
+		try {
+			var data = getVcsData(versionInfo);
+			return VcsInstall(data.type, data.data);
+		} catch (e:String) {
```
Simn, line 62: I don't like how we catch a string exception and then ignore it by throwing a different string exception. I think functions like `getVcsData` should just return an `Option` in cases like that. But yes I know you didn't write this, so we can change this later.

src/haxelib/api/Connection.hx
tobil4sk: The `haxelib.api` package should be fine, it's just the top level `haxelib` package where things have to be compilable with haxe 3. The `keyValueIterator` method for this abstract is safely kept behind conditional compilation.

src/haxelib/VersionData.hx
tobil4sk: Ah I probably did write this, just at an earlier point...

tobil4sk: I'll change it because this is the only place where the function is called now, so shouldn't be too hard to refactor.

src/haxelib/api/GlobalScope.hx
```diff
@@ -186,15 +187,18 @@ class GlobalScope extends Scope {
...
 
 			while (dependencies.length > 0) {
 				final dependency = dependencies.pop();
 				stack.add({
 					library: ProjectName.ofString(dependency.name),
 					version: // TODO: maybe check the git/hg commit hash here if it's given?
-						if (dependency.version == DependencyVersion.DEFAULT) null
-						else Version.ofString(dependency.version)
+						switch dependency.versionData {
+							case None: null;
```
Simn, line 198: Hmm, most of the recent changes went from `null` stuff to `Option` stuff, but here it's the other way around. If `version` is optional, we might want to turn that into an Option as well. But that's also not very urgent.

Simn:
This whole `hg` part isn't really a strong selling point for a PR, but this looks like good cleanup in general, so I'm happy to merge this change.

tobil4sk:
Yes, the code for the functionality existed anyway, it's just that the code for parsing `-lib` parameters and dependency strings was in different places and the latter was missing this for some reason. So it's really a side effect of cleaning up the duplicate functionality.

PR #559, opened 2022-04-14
Show allowed licenses if license is invalid

tobil4sk:
`Error: Invalid value <blah blah> for License.` isn't a very useful error message.

Now shows the user the permitted values, as before:
```Error: Invalid value `<blah blah>` for License. Allowed values: GPL, LGPL, MIT, BSD, Public, Apache```


Simn:
While we're here, what does it say if the licence is missing completely?

tobil4sk:
`Error: missing field license`

PR #560, opened 2022-04-21
Fix run with old neko versions

tobil4sk:
Neko 2.3.0 and earlier had a bug which prevented unsetting variables


tobil4sk:
Fixes https://github.com/HaxeFlixel/flixel/issues/2534

Issue #511, opened 2021-04-17
When installing default libs (eg. hxcpp) check that lib isn't already specified


haxiomic:
I need to use a specific version of hxcpp, so I add `--library hxcpp:git:https://github.com/haxiomic/hxcpp#patch-3` to the hxml

when users run `haxelib install build.hxml`. The result is both hxcpp 4.2.1 and git are installed but 4.2.1 takes priority:

```
Loading info about the required libraries
Haxelib is going to install these libraries:
  hxcpp
  haxe-c-bridge - 0.5.3
  hxcpp - 4.2.1
Continue? [y/n/a] ? y
```

haxiomic:
Hope to PR a fix myself, not a major change

Simn:
I got a request from a Haxe Foundation partner to prioritize this issue. So, any chance for that PR? :)

haxiomic:
It’s crunch time this week for me on contract work :(:(, I can get to this but after Nov 19th

RealyUniqueName:
I've pushed a fix, but it needs a test to close the issue.

tobil4sk:
I've looked into this a bit, and it seems to be completely unrelated from the fact that it is a "default lib". The issue arrises from doing this:
```hxml
-lib heaps
-lib heaps:1.8.0
```
This is because in the entry in the map for the first one is "heaps", but for the second one it's "heaps:1.8.0", so they don't get treated as the same library as they should.

tobil4sk:
#530 fixes this, and includes a test.

tobil4sk:
Apologies, I misunderstood the problem because I couldn't reproduce it (the -cpp flag has to be in a separate hxml). Here's the test for it: #532.

tobil4sk:
> I've pushed a fix, but it needs a test to close the issue.

Should be good to close now that the test is done, right?

PR #569, opened 2022-10-06
Fix minor issues relating to #511

tobil4sk:
This fixes two main issues with installing from hxml (#511):

- If a target library is specified to be installed from git, then we don't need to check the haxelib server's available versions, as we're not installing from there anyway. (This will also avoid the failing test in #564)
- We ignore target compilation flags (e.g. --cpp, --java, etc.) if the target libraries are already specified, regardless of the capitalisation of the library name. So that `-lib HXCPP` will still prevent the duplicate install.


src/haxelib/api/LibFlagData.hx
```diff
@@ -63,9 +63,13 @@ private final libraryFlagEReg = ~/^-(lib|L|-library)\b/;
...
 	Does not filter out repeated libs.
+
+	Target libraries (e.g. hxcpp) are also added if they are not explicitly
+	included via a -lib flag.
  **/
-function fromHxml(path:String):List<{name:ProjectName, data:Option<VersionData>, isTargetLib:Bool}> {
-	final libsData = new List<{name:ProjectName, data:Option<VersionData>, isTargetLib:Bool}>();
+function fromHxml(path:String):List<{name:ProjectName, data:Option<VersionData>}> {
+	final libsData = new List<{name:ProjectName, data:Option<VersionData>}>();
+	final targetLibraries = [for (_ => lib in TARGETS) lib => null];
```
Simn, line 72: Does this infer correctly on Haxe 4.2? I'm not sure if this ends up being `Null<Bool>` there, which would make this tri-state not work. Makes no difference on neko, but could be relevant if we ever use haxelib on a static target.

tobil4sk: Good catch.

Issue #567, opened 2022-10-01
Clarify purpose of user's email during registration


tobil4sk:
When a user is creating an account, we should make it clear that the email they give will be displayed publicly. Right now, the prompt just says:
```
Email:
```

Also, I don't think there is a check to make sure that emails are unique.

Simn:
Where are we showing email addresses publicly? Because I don't think we should do that...

tobil4sk:
Whenever someone runs:
```
haxelib user ...
```

To be honest, if this is the only place, we could just change this on the server right now and have it send an empty string instead of the actual email whenever someone runs `haxelib user`. I just assumed this was there so that people could contact library authors, but I guess if we're gonna allow that, it should probably be made optional anyway.

tobil4sk:
@Simn So it looks like it should be simple enough to patch this. Old clients are still going to show a "Mail: " line when running `haxelib user` however. So we can either send `null` in place of the email address, or something more descriptive so it's clear that it wasn't an error, like maybe `(private)` or something like that.

Simn:
Sounds good to me!

tobil4sk:
Should we also handle this for the legacy server? I know technically it's meant to be left as is, but perhaps it makes sense to change it in this case. Also it might be good for someone who knows more about the website to double check whether the email is not available in any other way. The email seems to be hashed with md5 when getting a list of users for the users page, I'm not sure if that ever ends up getting sent to the client.

tobil4sk:
The server seems to be still sending the email addresses with `haxelib user`. Does the server need to be updated manually @andyli?

andyli:
The `development` branch here get deployed to https://development-lib.haxe.org/. You may test it with `haxelib -R https://development-lib.haxe.org ...` first. If everything's alright, merge into `master` and it will be deployed to https://lib.haxe.org/.

tobil4sk:
Ah, I see. That is useful to know. This change is really small, and just substitutes the user's email for a "(private)" string. It seems to work:
```sh
> haxelib -R https://development-lib.haxe.org user ncannasse
Id: ncannasse
Name: Nicolas Cannasse
Mail: (private)
Libraries: 
  format
  dbadmin
  hscript
  ...
```

However, it seems like there are other unmerged changes in development that are missing from master. Is there anything else that should be checked before it is safe to merge?

andyli:
[...]

PR #571, opened 2022-10-07
Make docs compatible with old client versions

tobil4sk:
This changes the docs so that they mention the version from which certain bug fixes are available.


tobil4sk:
@Simn Do the documentation changes here look good to merge? They're required to advance with #567.

tobil4sk:
I'll merge this for now, we can always come back and change this if there are any issues.

PR #579, opened 2022-12-28
fix(schema): Add title to schema

hyperupcall:
As a part of restructuring in the SchemaStore repository, I noticed that a `title` property was defined in [the schemastore haxelib.json](https://github.com/SchemaStore/schemastore/blob/fa3112f1fccd1a50cf9242342d154a4103972c5a/src/schemas/json/haxelib.json), but not in this one. I thought it would be good to add it here to ensure that no information is lost during the restructuring (it will be removed).


tobil4sk:
Thanks!

PR #592, opened 2023-03-02
Print haxelib version used to stderr instead of stdout

tobil4sk:
See https://github.com/HaxeFoundation/haxelib/pull/555#issuecomment-1450521837.

Lime sometimes runs `haxelib path --debug ...` so having this in stdout breaks builds.


tobil4sk:
@Simn If you don't have any objections, I'll go ahead and merge this.

Simn:
I always forget how Haxe and haxelib communicate, but if this doesn't get in the way then I have no objections.

PR #597, opened 2023-04-01
Perform haxelib validation on server side as well

tobil4sk:
The client may be outdated or otherwise modified, so we cannot trust that it will always send valid libraries


tobil4sk:
I've added some missing tests for the dependency checks. This should be safe to merge since it just uses the already existing checks.

Issue #591, opened 2023-03-01
Git update sometimes crashes after #585


tobil4sk:
Occasionally when running `haxelib update` for a git repository (on Windows), there is a crash. 

<details>
  <summary>Crash logs</summary>

```
$ haxelib update hxcpp --debug
[debug] Using haxelib from "..."
# Running command: git [diff,--exit-code,--no-ext-diff]

# Exited with code 0

# Running command: git [diff,--cached,--exit-code,--no-ext-diff]

# Exited with code 0

# Running command: git [fetch]

# Exited with code 0

# Running command: git [rev-parse,@{u}]

3f6de84d4decb0a7aa1131ebd527f623b1e2d2b1

# Exited with code 0

# Running command: git [rev-parse,HEAD]

An exception occurred in a neko Thread :
std@lock_release
An exception occurred in a neko Thread :
std@lock_release
Called from ? line 1
Called from haxelib/Util.hx line 14
Called from haxelib/client/Main.hx line 895
Called from haxelib/Util.hx line 14
Called from haxelib/client/Main.hx line 268
Called from haxelib/client/Main.hx line 583
Called from haxelib/api/Installer.hx line 372
Called from haxelib/api/Installer.hx line 373
Called from haxelib/api/Installer.hx line 426
Called from haxelib/Util.hx line 14
Called from haxelib/api/Installer.hx line 826
Called from haxelib/api/Vcs.hx line 288
Called from haxelib/api/Vcs.hx line 167
Called from haxelib/api/Vcs.hx line 206
Called from C:\HaxeToolkit\haxe\std/neko/_std/sys/io/Process.hx line 108
Uncaught exception - std@process_exit
```

```
$ haxelib update hxcpp --debug
[debug] Using haxelib from "..."
# Running command: git [diff,--exit-code,--no-ext-diff]

# Exited with code 0

# Running command: git [diff,--cached,--exit-code,--no-ext-diff]

# Exited with code 0

# Running command: git [fetch]

# Exited with code 0

# Running command: git [rev-parse,@{u}]

3f6de84d4decb0a7aa1131ebd527f623b1e2d2b1

# Exited with code 0

# Running command: git [rev-parse,HEAD]

An exception occurred in a neko Thread :
std@lock_release
An exception occurred in a neko Thread :
std@lock_release
Called from ? line 1
Called from haxelib/Util.hx line 14
Called from haxelib/client/Main.hx line 895
Called from haxelib/Util.hx line 14
[...]

PR #607, opened 2023-07-10
Optimize git library installation

### LINKED ISSUE #493: haxelib git without .git folder?
Is it possible to install haxelib with "haxelib git " command, but do not clone .git folder, that contains all history?

### LINKED ISSUE #512: Haxelib re-installs git libs despite saying they're already installed
After `haxelib install build.hxml`, it skips already installed libraries, unless those are git libs, in which case it installs them again:

```
You already have hxcpp version git installed.
Installing hxcpp from https://github.com/haxiomic/hxcpp branch: patch-3
Library hxcpp current version is now git
```

This can be slow when you just want to insure the dependencies are correct before running `haxe build.hxml`

I intend to PR a fix (no timeline however)

### LINKED ISSUE #593: Git submodules can be checked out incorrectly
We clone the repository and update the submodules before checking out to the correct branch. This means the submodules are checked out to the versions in the default branch, which might be different from the ones in the new branch. If a branch/commit/tag is specified, we should delay the submodule update until after we check out to the correct commit.

tobil4sk:
Closes #493. For most git installs, a (faster) shallow clone can now be performed with `--depth=1`, which avoids cloning the entire history and instead only clones one commit's worth of history, including for the submodules. The only exception where a shallow clone cannot be performed is when setting a library to a specific commit with a short commit id, for example:
```sh
# Reverts to full non-shallow clone:
haxelib git lime https://github.com/openfl/lime c16f278
# Performs shallow clone:
haxelib git lime https://github.com/openfl/lime c16f27818dcff17e8b5e4648624a64ea6a107119
```
This is due to how commit ids are resolved.

The data which was used to install the library is now stored in a `.gitdata` file, to avoid issues of overwriting externally created commits.

Now when updating a git library, it will now only update if it was installed with the default branch or a specific branch. If it was installed with a specific commit or tag, it remains at that commit/tag, since there is no clear commit to update to.

Mercurial works differently, and doesn't have a concept of a shallow clone so I've just left it more or less as is.

### DIFF
--- a/src/haxelib/VersionData.hx
+++ b/src/haxelib/VersionData.hx
@@ -19,6 +19,13 @@ package haxelib;
 		else
 			throw 'Invalid VscID $s';
 	}
+
+	public function getName() {
+		return switch cast(this, VcsID) {
+			case Git: "Git";
+			case Hg: "Mercurial";
+		};
+	}
 }
 
 /** Class containing repoducible git or hg library data. **/
--- a/src/haxelib/VersionData.hx
+++ b/src/haxelib/VersionData.hx
@@ -28,7 +28,7 @@ class VcsData {
 	var url:String;
 	/** Commit hash **/
 	@:optional
-	var ref:Null<String>;
+	var commit:Null<String>;
 	/** The git tag or mercurial revision **/
 	@:optional
 	var tag:Null<String>;
@@ -45,7 +45,7 @@ class VcsData {
 
 	public function toString(): String {
 		var qualifier =
-			if (ref != null) ref
+			if (commit != null) commit
 			else if (tag != null) tag
 			else if (branch != null) branch
 			else null;
@@ -89,7 +89,7 @@ class VersionDataHelper {
 			type: type,
 			data: {
 				url: vcsRegex.matched(2),
-				ref: vcsRegex.matched(3),
+				commit: vcsRegex.matched(3),
 				branch: vcsRegex.matched(4),
 				subDir: null,
 				tag: null
--- a/src/haxelib/VersionData.hx
+++ b/src/haxelib/VersionData.hx
@@ -62,13 +62,17 @@ class VcsData {
 			url;
 	}
 
+	/**
+		Returns whether this vcs data will always reproduce an identical installation
+		(i.e. the commit id is locked down)
+	**/
 	public function isReproducible() {
 		return commit != null;
 	}
 
 	/**
-		Returns an object containing the filled-in VcsData fields,
-		without the empty ones.
+		Returns an anonymous object containing only the non-null, non-empty VcsData fields,
+		excluding the null/empty ones.
 	 **/
 	public function getCleaned() {
 		var data:{
--- a/src/haxelib/VersionData.hx
+++ b/src/haxelib/VersionData.hx
@@ -61,6 +61,35 @@ class VcsData {
 		else
 			url;
 	}
+
+	public function isReproducible() {
+		return commit != null;
+	}
+
+	/**
+		Returns an object containing the filled-in VcsData fields,
+		without the empty ones.
+	 **/
+	public function getCleaned() {
+		var data:{
+			url:String,
+			?commit:String,
+			?tag:String,
+			?branch:String,
+			?subDir:String
+		} = { url : url };
+
+		if (commit != null)
+			data.commit = commit;
+		if (tag != null)
+			data.tag = tag;
+		if (!(branch == null || branch == ""))
+			data.branch = branch;
+		if (!(subDir == null || haxe.io.Path.normalize(subDir) == ""))
+			data.subDir = subDir;
+
+		return data;
+	}
 }
 
 /** Data required to reproduce a library version **/
--- a/src/haxelib/api/GlobalScope.hx
+++ b/src/haxelib/api/GlobalScope.hx
@@ -63,7 +63,7 @@ class GlobalScope extends Scope {
 			repository.setVcsData(library, vcsVersion, data);
 		}
 
-		if (!(data == null || data.subDir == "" || data.subDir == null)) {
+		if (data != null && data.subDir != "" && data.subDir != null) {
 			final devDir = repository.getValidVersionPath(library, vcsVersion) + data.subDir;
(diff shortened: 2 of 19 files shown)


Simn:
@kLabz Could you check if this makes sense git-wise?

kLabz:
I don't think I'll have time today, so will look at this tomorrow.

It likely breaks some workflow I use from time to time (using git dependency, and doing git commands in `.haxelib/somelib/git` folder like changing branches etc.) but I should probably use haxelib dev for that as it's easy to lose changes like that...

tobil4sk:
Somehow this PR causes #591 to be triggered on Windows with pretty much every `haxelib git` command :/ Probably requires getting rid of neko to fix it, I can't figure out what exactly is going on...

dpomier:
> The only exception where a shallow clone cannot be performed is when setting a library to a specific commit with a short commit id

It should be possible to get the full commit id from the short one once you have added a git remote: 
```
$ git rev-parse 3cdd5d
3cdd5d19178a54d2e51b5098d43b57571241d0ab
```

> Now when updating a git library, it will now only update if it was installed with the default branch or a specific branch. If it was installed with a specific commit or tag, it remains at that commit/tag, since there is no clear commit to update to.

It might be worth always fetching the commit id so as not to lose your optimisation when a branch or tag is specified instead of a commit id. This can be done this way:
```
$ git ls-remote https://github.com/HaxeFoundation/haxelib.git HEAD     
98637027327d8cf385d302acaaf104bd6107d2bf        HEAD

$ git ls-remote https://github.com/HaxeFoundation/haxelib.git 4.1.x
f17fffa97554b1bdba37750e3418051f017a5bc2        refs/heads/4.1.x
```
Hope this helps.

tobil4sk:
> It should be possible to get the full commit id from the short one once you have added a git remote:

This doesn't work if the commit hasn't been fetched:
```
$ git rev-parse 558fd07 
558fd07
fatal: ambiguous argument '558fd07': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

> It might be worth always fetching the commit id so as not to lose your optimisation when a branch or tag is specified instead of a commit id.

Not sure I understand, which optimisation are you talking about that would be lost?

tobil4sk:
The main issue that's blocking this is #591, there is no point in optimising git commands if it will break them :sweat_smile:. Maybe it can work in development with the interp target and hxp, but that would mean that it doesn't work with haxe 4.3.x.

tobil4sk:
The cause for the crashes has been resolved now! I've updated the branch to resolve conflicts, and I've also split out the commits into more atomic changes and added more descriptive messages (would be good to rebase rather than squash to preserve history). This is now ready for review.

Mercurial now also is consistent with Git's behaviour.

I've also taken out the changes for #512, which will come in a separate PR.

I also had to disable the cpp client tests, because default arguments in abstract classes are broken on cpp, see: https://github.com/HaxeFoundation/haxe/issues/11666.

PXshadow:
Tested on go2hx git library with help from @tobil4sk . It's over 200MB if not shallow cloned, and with 3 different git submodules. The time it took to install from git in total went from 14 seconds to 3 seconds on my machine, using this branch.

src/haxelib/api/GlobalScope.hx
```diff
@@ -59,9 +59,11 @@ class GlobalScope extends Scope {
 	}
 
 	public function setVcsVersion(library:ProjectName, vcsVersion:VcsID, ?data:VcsData):Void {
-		if (data == null) data = {url: "unknown"};
+		if (data != null) {
+			repository.setVcsData(library, vcsVersion, data);
+		}
 
-		if (data.subDir != null) {
+		if (!(data == null || data.subDir == "" || data.subDir == null)) {
```
Simn: This looks weird, could we De Morgan it?

test/RunCi.hx
```diff
@@ -385,7 +385,7 @@ Listen 2000
 			"-D",
 			'haxelib_path=${Path.join([Sys.getCwd(), "haxelib"])}'
 		]);
-		runCommand("neko", ["bin/integration_tests.n"]);
+		// runCommand("neko", ["bin/integration_tests.n"]);
```
Simn, line 388: Is this supposed to be disabled now?

src/haxelib/VersionData.hx
```diff
@@ -54,6 +61,35 @@ class VcsData {
 		else
 			url;
 	}
+
+	public function isReproducible() {
```
Simn, line 69: Some documentation would be good here because I don't know what this means.

src/haxelib/VersionData.hx
```diff
@@ -54,6 +61,35 @@ class VcsData {
...
+		return commit != null;
+	}
+
+	/**
+		Returns an object containing the filled-in VcsData fields,
+		without the empty ones.
```
Simn: Uhm, what? I struggle to make sense of both this description and the implementation. What is the point of this function?

src/haxelib/api/GlobalScope.hx
tobil4sk: Updated

test/RunCi.hx
tobil4sk: Yes they should be for now, it is explained in the commit message of 824c06d080aa885021273acc33af574e26a539a8. This test run uses the cpp binary as the haxelib client (specified by `haxelib_path`), but that is currently broken. The test with the regular neko client still runs on the lines above.

src/haxelib/VersionData.hx
tobil4sk: It converts a VcsData class object into an anonymous object with the fields that actually carry any information. This is useful when you want to write to a json file, where including the null fields is pointless.

This is haxelib's cli for git installs:
```
haxelib git [project-name] [git-clone-path] [branch] [subDir] [tag]
```

So if you want to install with a tag but not a branch or subDir, you have to put both of those anyway as empty strings. The method also excludes fields that have empty values like this, which don't actually hold any useful information about the installation.

PR #610, opened 2023-10-12
Add "state save" and "state load" commands

j-jzk:
This PR adds two new commands:
* `haxelib state save haxelib-lock.hxml`
* `haxelib state load haxelib-lock.hxml`

This manifest file (note that `.hxml` extension isn't mandatory here) is used to "lock" the current state of the libraries in the project. This is useful for reproducing the exact state of dependencies in a project (on a new machine or in Docker), especially when working with libraries installed from git.

The intended workflow is following:
1. do package management operations
2. lock the dependencies: `haxelib state save haxelib-lock.hxml`
3. apply the manifest file in another environment: `haxelib state load haxelib-lock.hxml`

This manifest mechanism supports published haxelib versions, as well as exact git and mercurial references. `haxelib dev` libs are ignored (with a warning), and try to resolve to "non-dev" version for the manifest file, allowing you to work with dev libs locally without altering the pin file.

I switched to file output instead of stdout for better cross-platform support, and API consistency.
"manifest" terminology comes from https://learn.microsoft.com/en-us/vcpkg/users/manifests

---
Original PR description below:

This PR adds a new _lock_ command which, as in other library managers, "locks" the current state of the libraries in the project. This is useful for reproducing the exact state of dependencies in a project (on a new machine or in Docker), especially when working with libraries installed from git.

The output is in HXML (`-L lib:version ...`) and it is dumped to stdout instead of a file.
The intended workflow is following:
1. do package management operations
2. lock the dependencies: `haxelib lock > haxelib-lock.hxml`
3. apply the lockfile in another environment: `haxelib install --skip-dependencies --always haxelib-lock.hxml`

I've created this for my personal usage and providing this patch hoping it could be useful to more developers. If inappropriate, feel free to add notes/close the PR.
Before merging, support for locking versions of Mercurial libs (installed using `haxelib hg ...`) should be added. I am not a Mercurial user myself, so help would be greatly appreciated.


src/haxelib/client/Main.hx
```diff
@@ -843,6 +846,39 @@ class Main {
...
+
+		// sort projects alphabetically
+		libraryInfo.sort(function(a, b) return Reflect.compare(a.name.toLowerCase(), b.name.toLowerCase()));
+
+		for (library in libraryInfo) {
+			if (library.devPath != null) {
+				Cli.printError('Error: version of "${library.name}" is "dev". The lockfile will not work properly in other environments!');
+			}
+
+			final version = scope.getVersion(library.name);
```
kLabz, line 866:
```diff
-                       final version = scope.getVersion(library.name);
+                       final version = try scope.getVersion(library.name) catch (e) {
+                               Cli.printError('Error: ${e.message}');
+                               continue;
+                       };
```

Dev libs that don't have "non dev" version installed will make `haxelib lock' abort early if you don't catch errors there.

src/haxelib/client/Main.hx
```diff
@@ -843,6 +846,39 @@ class Main {
...
+				final vcsId = VcsID.ofString(version);
+				versionData = VcsInstall(
+					vcsId,
+					scope.repository.getVcsData(library.name, vcsId)
+				);
+			} else {
+				versionData = Haxelib(SemVer.ofString(version));
+			}
+			final versionStr = VersionDataHelper.toString(versionData);
+			Cli.print('-L ${library.name}:$versionStr');
```
kLabz, line 878: I'm still not sure about generating `-L` instead of `--lib` because there are still people dealing with older versions. Might be just me, though..

src/haxelib/client/Args.hx
```diff
@@ -255,6 +257,7 @@ class Args {
 		addCommand(Update, "update a single library (if given) or all installed libraries", Basic);
 		addCommand(Remove, "remove a given library/version", Basic);
 		addCommand(List, "list all installed libraries", Basic);
+		addCommand(Lock, "output an hxml file with all the installed libraries", Basic);
```
kLabz, line 260: I think I'd move that command from `Basic` into `Development` category

src/haxelib/api/Vcs.hx
```diff
@@ -347,6 +350,20 @@ class Git extends Vcs {
 		// return prev. cwd:
 		Sys.setCwd(oldCwd);
 	}
+
+	public function getReproducibleVersion(libPath: String): VcsData {
+		final oldCwd = Sys.getCwd();
+		Sys.setCwd(libPath);
+
+		final ret: VcsData = {
+			// could the remote's name not be "origin"?
```
kLabz, line 359: Well yeah it could.. but "most of the time it will work all the time" :shrug: 

Might need some error handling to recover just in case?

src/haxelib/api/Vcs.hx
```diff
@@ -420,4 +437,9 @@ class Mercurial extends Vcs {
 		if (run(vcsArgs, debugLog).code != 0)
 			throw VcsError.CantCloneRepo(this, url/*, ret.out*/);
 	}
+
+	public function getReproducibleVersion(libPath: String): VcsData {
+		// TODO
+		throw new haxe.exceptions.NotImplementedException("Getting revision hash isn't supported for Mercurial");
```
kLabz, line 443: Ok so I was able to get an implementation:
```haxe
		final oldCwd = Sys.getCwd();
		Sys.setCwd(libPath);

		final ret: VcsData = {
			url: run(["paths", "default"], true).out.trim(),
			ref: run(["identify", "-i", "--debug"], true).out.trim(),
		};

		Sys.setCwd(oldCwd);
		return ret;
```
**But** that doesn't work for installation, since `-L foo:hg:...#rev` will try to install `rev` as being a branch exclusively.

I had to update `clone` too:
```diff
        public function clone(libPath:String, url:String, ?branch:String, ?version:String, ?debugLog:(msg:String)->Void):Void {
                final vcsArgs = ["clone", url, libPath];
 
-               if (branch != null) {
+               if (branch != null && version != null) {
                        vcsArgs.push("--branch");
                        vcsArgs.push(branch);
-               }
 
-               if (version != null) {
                        vcsArgs.push("--rev");
                        vcsArgs.push(version);
+               } else if (branch != null) {
+                       vcsArgs.push("--updaterev");
+                       vcsArgs.push(branch);
+               } else if (version != null) {
+                       vcsArgs.push("--updaterev");
+                       vcsArgs.push(version);
                }
 
                if (run(vcsArgs, debugLog).code != 0)
                        throw VcsError.CantCloneRepo(this, url/*, ret.out*/);
        }
```

But I'd like to add proper tests for that. Only tested with `https://www.mercurial-scm.org/repo/hello` locally with:
```sh
haxelib hg test-tag https://repo.mercurial-scm.org/hello tip
haxelib hg test-branch https://repo.mercurial-scm.org/hello default
haxelib hg test-branch-tag https://repo.mercurial-scm.org/hello default "" tip
haxelib hg test-rev https://repo.mercurial-scm.org/hello 82e55d328c8ca4ee16520036c0aaace03a5beb65
```

Which all worked

kLabz:
Thanks!
I think this can be merged once the above points have been handled

src/haxelib/client/Main.hx
j-jzk: People using `haxelib lock` won't be using versions not supporting `-L` :)
I will change it to `--lib` anyway for clarity.

kLabz: > People using haxelib lock won't be using versions not supporting -L :)
I disagree. I would use haxelib lock but I also support CI for multiple haxe versions.

j-jzk:
Hopefully I will finish this till the end of the week.

kLabz:
This PR is now ready as far as I'm concerned.
Will wait a bit before merging in case someone has some argument against it.

kLabz (approved):

src/haxelib/client/Args.hx
```diff
@@ -276,6 +278,7 @@ class Args {
 		addCommand(Setup, "set the haxelib repository path", Miscellaneous);
 		addCommand(NewRepo, "create a new local repository", Miscellaneous);
 		addCommand(DeleteRepo, "delete the local repository", Miscellaneous);
+		addCommand(Lock, "output an hxml file with all the installed libraries", Miscellaneous);
```
Simn, line 281: This doesn't seem like the most relevant description...

src/haxelib/client/Main.hx
```diff
@@ -843,6 +846,43 @@ class Main {
 		Cli.print('Local repository deleted ($path)');
 	}
 
+	/** Outputs a lock file with all the installed libraries **/
+	function lock() {
+		if (RepoManager.getLocalPath(Sys.getCwd()) == null) {
+			Cli.printWarning("no local repository found, dumping lock from global");
```
Simn, line 852: wtf does "dumping lock from global" mean?

src/haxelib/client/Main.hx
```diff
@@ -843,6 +846,43 @@ class Main {
...
+
+		final scope = getScope();
+		final libraryInfo = scope.getArrayOfLibraryInfo();
+
+		// sort projects alphabetically
+		libraryInfo.sort(function(a, b) return Reflect.compare(a.name.toLowerCase(), b.name.toLowerCase()));
+
+		for (library in libraryInfo) {
+			if (library.devPath != null) {
+				Cli.printError('Error: version of "${library.name}" is "dev". The lockfile may not work properly in other environments!');
```
Simn, line 863: If this is an error, does it mean I cannot use the command at all as soon as I have a single dev library? That seems rather limiting... Shouldn't it rather warn about this and ignore dev libraries?

kLabz: Ah, right I wanted to use "Warning" wording there.

I do want only warnings because I often have lock files that I want to keep working, but while actually working locally I also often work on some of these libs, so I use `haxelib dev` on them.

Current behavior warns about dev libs being present and so their line in output being potentially outdated, but still goes on.

src/haxelib/client/Args.hx
kLabz: Is "output an hxml file with installation instructions for installed libraries" any better?

Simn: Uh, barely...

I'd expect a lock command to mention, you know, locking? The fact that HXML files are involved doesn't even seem particularly important.

kLabz: Well that's the thing, it doesn't lock anything. Maybe the issue is more with the command name?

Simn: Are you saying we have a lock command that creates a lock file that doesn't lock anything? I'm not sure I understand what's happening here...

Simn: Oh so that's what happens here... in this case I agree that "lock" is not a good name, and the description should probably end with "for use with `haxelib install`".

kLabz: Hopefully this version is better (check updated PR description)

j-jzk:
Thanks for all the work :) Just sat down to work on this and saw you'd already finished it (much better than I would have).

kLabz:
Hope it still matches your workflow! Changed into a file arg so that windows people don't freak out :sweat_smile:

kLabz:
@Simn is it ok to merge this?
Documentation will wait a bit (for a new haxelib release), and can be done on a separate PR.

PR #619, opened 2023-12-27
Add server-side hashing of passwords

tobil4sk:
Implements argon2id hashing of passwords on the server, with automatic rehashing of the old Md5 hashes. The Md5 hashing on the client is kept as is for now.

See #314.


src/haxelib/server/Update.hx
```diff
@@ -0,0 +1,61 @@
+package haxelib.server;
+
+import haxelib.server.SiteDb;
+
```
andyli, line 4: I think we need to use transactions for the public methods of the `Update` class. There are mulpile replicas of haxe lib server pod running in the cluster.

andyli: Also, both dev (development-lib.haxe.org) and prod (lib.haxe.org) connect to the same DB. When dev updated the DB, would prod (that runs older code) still works correctly with the upgraded DB? I haven't think through all the features yet, but e.g. registering new user may not work because the `salt` column is `NOT NULL`.

tobil4sk: > I think we need to use transactions for the public methods of the Update class. There are mulpile replicas of haxe lib server pod running in the cluster.

I've wrapped the methods in transactions. It looks like the `Manager.all()` method uses locking by default, though this only works on mysql, not sqlite, though it seems the server is using mysql anyway?

https://github.com/HaxeFoundation/record-macros/blob/00afa3924e8478203dd500f8b8a6f2174b2f8e13/src/sys/db/Manager.hx#L475-L477
https://github.com/HaxeFoundation/record-macros/blob/00afa3924e8478203dd500f8b8a6f2174b2f8e13/src/sys/db/Manager.hx#L54-L56

> When dev updated the DB, would prod (that runs older code) still works correctly with the upgraded DB?

This would be an issue. Would it be possible to disable the development server while this transition is taking place? Otherwise, I can open a PR to temporarily disable any registrations or other commands that would be a problem. Then we can push it to disable that on the prod server temporarily, perform the migration, and then re-enable everything once it is done.

src/haxelib/server/SiteDb.hx
```diff
@@ -145,11 +151,25 @@ class SiteDb {
...
-		for (m in managers)
-			if (!TableCreate.exists(m))
+
+		var hasOldTables = false;
+
+		for (m in managers) {
+			if (TableCreate.exists(m)) {
+				hasOldTables = true;
+			} else {
 				TableCreate.create(m);
```
tobil4sk, line 164: @andyli Regarding what you said about multiple server instances, this may be a problem here since checking if a table exists and creating it is not done atomically. I'm not sure if this might end up being an issue if two instances both add conflicting `Meta` table entries, which is meant to be a table with a single field? If the Table.create would fail in that case anyway then I guess it wouldn't be a problem.

andyli: Yeah. I think the latter command will fail when there are two parallel `TableCreate.create()`. Ideally we should use `CREATE TABLE IF NOT EXISTS`... but well.

I agree it should be fine to let the latter one fail since apache/k8s will restart the instance.

src/haxelib/server/Update.hx
tobil4sk: I have updated the PR to implement the first step of the migration as described in your comment.

src/haxelib/server/SiteDb.hx
```diff
@@ -32,6 +32,8 @@ class User extends Object {
 	public var fullname : String;
 	public var email : String;
 	public var pass : String;
+	public var pass2 : Null<String>;
+	public var salt : Null<SBytes<16>>;
```
andyli: Looks like `salt` is redundant because the hash in `pass2` already contains the salt?

Also, `generateSalt()` returns a 32-byte salt but here the type is 16.

tobil4sk: Yes, the salt field was never actually read (except for a test, which was redundant anyway). I pushed an updated patchset without it.

PR #625, opened 2024-04-09
Exit with error when hxml install fails

kLabz:
One thing that bothers me a lot with CI (along with windows CI often ignoring failures..) is haxelib install failures (mostly due to network issues) being silently ignored because haxelib happily returns a 0 exit code even when some (or all) libs couldn't be installed.


Simn:
Good idea, this has bothered me too.

kLabz:
And we get the main thing I dislike with CI: earthly failures :sweat_smile:

PR #630, opened 2024-05-24
Add MIT license to haxelib

Maniues:
This patch adds MIT license to this repo (as stated in package.json and haxelib.json).

You can modify this license to add to copyright notice other contributors included in haxelib.json.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2024 Haxe Foundation
+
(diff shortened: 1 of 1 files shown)


Simn:
I have some vague memory of the licensing being a bit more difficult for haxelib in particular, but I don't remember any details.

Simn:
Ah okay, if the haxelib.json already says it then I guess somebody has thought about it before.

PR #634, opened 2024-07-30
Display current scope when using `haxelib list`

ninjamuffin99:
Adds a file system check for a `.haxelib` directory in `GlobalScope.hx` (there's no `LocalScope.hx` class), and then sets `isLocal` accordingly.

Then in the `Main.list()` function, we check for `isLocal` being true or false. We display `Local Haxelib Repository at: {localPath}` if we are in local scope, and `Global Haxelib at: ${RepoManager.getGlobalPath()}\n`

also slightly change the way that list works, appending just one string and printing a single string to the CLI, rather than printing each string individually

*In the root of FNF (.haxelib directory exists)*
image

*In the `assets/` folder/submodule of FNF (.haxelib directory does not exist*
image


EliteMasterEric (approved):

kLabz:
This needs to handle the case where global scope doesn't exist.

kLabz (changes requested):
Local scope is not only for when there's a `.haxelib` folder in current repository (can be in any parent directory, too), and in this case:
* this PR considers the repo as global while it's not
* it breaks if global scope isn't setup

ninjamuffin99:
> This needs to handle the case where global scope doesn't exist.

just double checking, global scope is where haxelibs get installed to, and gets set up via `haxelib setup`?

kLabz:
Yes. But you can also only ever use local repositories and never setup global repository to avoid mistakenly installing libs globally.

ninjamuffin99:
nice, just made change that uses `RepoManager.getPath()` which recursively looks for a local haxelib repository. This removes my old method of looking for `.haxelib` specifically.

Then it compares whatever path it finds to the global repository path via `RepoManager.getGlobalPath()`. If it matches, it assumes that it's not local. 

I'll now poke around into making sure that it works without a global scope setup!

src/haxelib/api/GlobalScope.hx
```diff
@@ -21,6 +23,9 @@ using StringTools;
 class GlobalScope extends Scope {
 	function new(repository:Repository) {
 		super(false, repository);
+		
+		// While this class is "GlobalScope", there's currently not a "LocalScope" equivalent class. For now we can add this here. 
+		isLocal = RepoManager.getPath() != RepoManager.getGlobalPath();
```
kLabz, line 28: Wouldn't `repository.path` work there?

ninjamuffin99: haha yes, I'm still learning my way around the haxelib codebase, will update it to that

ninjamuffin99:
so it's possible to have a local repo without having setup a global one? So this command should respond accordingly and work in local repos on machines with no global repo?

src/haxelib/api/GlobalScope.hx
```diff
@@ -21,6 +23,9 @@ using StringTools;
 class GlobalScope extends Scope {
 	function new(repository:Repository) {
 		super(false, repository);
+		
+		// While this class is "GlobalScope", there's currently not a "LocalScope" equivalent class. For now we can add this here. 
+		isLocal = repository.path != RepoManager.getGlobalPath();
```
kLabz, line 28: Hmm, `RepoManager.getGlobalPath()` will throw if there's no global repository setup.

This not so pretty hack might not be a good idea..
`isLocal` should probably live in `Repository` itself, which could also more easily identify local / global cases (might need to rework `get()` a bit there).

ninjamuffin99:
some changes
- moved the `isLocal` set code from `GlobalScope.hx` to `Scope.hx`
- removed the `isLocal` arg from `new Scope()`, since we determine if it's local or not depending on the directory information

When running `haxelib list` with no global scope, it behaves as normal, telling you that you should run `haxelib setup`. But if you have no global scope but run `haxelib list` inside a local repository (or a recursive subdirectory of a local repository) it will properly list out the local repositorys haxelibs

ninjamuffin99:
took a peak at some of those failed tests just now. 
- `TestDev.testPreferenceOfHaxelibJsonName()` wants the output to start with the name of the library 
- `TestList.testInvalidDirectories()` expects nothing in the output

Not to schmooze around the tests too much, what are some options to get these passing? I will say, I don't know testing etiquette! Off the top of my head, maybe can pass `--quiet` to these specific tests, and when `--quiet` gets passed through, it suppresses the `Repository at: {path} ` message, which maybe can be helpful if people want to do parsing on their `haxelib list` output in general.

kLabz:
Yeah, haxelib list parsing is one of my concerns, too.

ninjamuffin99:
- passing through `--quiet` to `haxelib list` will suppress the new line regarding current scope
- modified the failing tests to pass through `--quiet` as an arg, which will display the previous `haxelib list` "styling" (hopefully!)

tobil4sk:
> Adds a file system check for a .haxelib directory in GlobalScope.hx (there's no LocalScope.hx class), and then sets isLocal accordingly.

The `isLocal` member in the Scope class refers to a different concept than global/local repositories.

Repository - a place where libraries are installed, can hold multiple versions of the same library.
- global: all libraries are stored in a folder accessible from anywhere on the system.
- local: all libraries are stored in a `.haxelib` folder somewhere up the directory tree.

Scope - something takes a library name and tells you which version of that library to use.
- global: versions are specified in `.current` files (which currently happens to be in the repository).
- local: not implemented yet, but in future library versions may be stored in a project specific file, so that a project can specify which library versions it wants to use.

`isLocal` refers to global/local scopes, not global/local repositories.

ninjamuffin99:
ah that makes total sense. Knowing what I know now, I'll entirely ignore `isLocal`, and simply run the "local check" in the `list()` function itself, keeping `Scope.hx` and `GlobalScope.hx` untouched

PR #640, opened 2024-08-14
check for different urls when running `haxelib git`

ninjamuffin99:
Currently running `haxelib git` it doesn't check the differences in remote urls.

Running `haxelib git flixel https://github.com/HaxeFlixel/flixel.git` will correctly install `flixel`. However, if you then try running with a different url `haxelib git flixel https://github.com/FunkinCrew/flixel.git` it doesn't properly update / reinstall from the new URL. Going into the cloned repo `flixel/git` and running `git config --get remote.origin.url` will show `https://github.com/HaxeFlixel/flixel.git` rather than the new one.

This PR runs a check from the currently installed haxelib's git remote to the newly inputted one, and if it's different, it will ask if you want to install the new one. 

Right now I just have it implemented for git, not mecurial. Current implementation is agnostic to whether someone has an SSH remote url (`git@github.com:HaxeFlixel/flixel`) or HTTP remote url (`https://github.com/HaxeFlixel/flixel`)


src/haxelib/api/Vcs.hx
```diff
@@ -356,14 +358,39 @@ class Git extends Vcs {
...
 			ref: run(["rev-parse", "HEAD"], true).out.trim(),
 		};
 
 		Sys.setCwd(oldCwd);
 		return ret;
 	}
+
+	public function parseUrl(url:String):{protocol:String, domain:String, user:String, repo:String, cleanUrl:String}
+	{
+		var regex = new EReg("^(https:\\/\\/|git@)([^\\s/:]+)(?::|\\/)([^\\s/]+)\\/([^\\s/]+?)(?:\\.git)?$", "");
```
tobil4sk, line 375: I would prefer to avoid complicated parsing like this, especially because this will only work with github repository urls. I think it should be good enough to cover cases where the url matches identically.

ninjamuffin99: that sounds good, I suppose I was aiming to get it a bit cross compatible with ssh vs. git. If someone wants to download the git@ ssh version of a git repo, but their remote is already at the https version of a git repo, we might want to just update the remote rather than redownload. However just using identical urls sounds good

ninjamuffin99:
how was this one looking? Should I add tests for this?

tobil4sk:
> how was this one looking? Should I add tests for this?

Yes please, a test would be useful!

ninjamuffin99:
finally got around to figuring out how to draft up a test! 

I wrote it in `TestInstaller.hx` rather than `TestGit.hx` since the url checking logic was in `Installer.hx`.

src/haxelib/api/Installer.hx
```diff
@@ -795,8 +795,18 @@ class Installer {
 
 			final currentBranch = vcsBranchesByLibraryName[library];
 
-			// TODO check different urls as well
-			if (branch != null && (!wasUpdated || currentBranch != branch)) {
+			var libUrl = vcs.getReproducibleVersion(libPath).url;
+
+			if (url != null && libUrl != null && url != libUrl)
+			{
+				if (!userInterface.confirm('Update remote url: "${libUrl}" with "${url}"')) {
```
tobil4sk, line 802: This introduces a new cli interaction that wasn't present before, it may be best to simply overwrite unless there are uncommitted changes.

ninjamuffin99:
how does the PR look @tobil4sk ?

test/libraries/InstallDeps/git-deps-url-new.hxml
```diff
@@ -0,0 +1 @@
+-lib flixel:git:https://github.com/geokureli/flixel.git#4.5.0
```
tobil4sk, line 1: It should be possible to set up a url that is just a path to a local directory. There is a util function for setting up a git repository, and there is a `test/libraries` folder with some existing libraries that could be pointed to.

https://github.com/HaxeFoundation/haxelib/blob/5a836287828fdaeb6aa91695a5eb399cee0f6640/test/tests/util/Vcs.hx#L8

https://github.com/HaxeFoundation/haxelib/blob/5a836287828fdaeb6aa91695a5eb399cee0f6640/test/tests/integration/TestVcs.hx#L6-L8

External repositories may change ownership or move (and they rely on the network to clone), so it is more reliable to clone from local repositories in tests, especially the unit tests.

test/tests/TestInstaller.hx
```diff
@@ -105,6 +106,19 @@ class TestInstaller extends TestBase {
 		assertTrue(scope.isLibraryInstalled(ProjectName.ofString("hx.signal")));
 	}
 
+	public function testInstallUrlChanges() {
+		installer.installFromHxml("git-deps-url-old.hxml");
```
tobil4sk, line 110: For this test, we should run `installVcsLibrary` rather than `installFromHxml`, since we are only testing the git functionality

tobil4sk (changes requested):

PR #641, opened 2024-08-17
Show progress in git based installs

### LINKED ISSUE #635: Display progress for Git based installs

ninjamuffin99:
closes #635 

Adds a `--progress` flag to the `git clone` command, which will spit out the information / progress of the cloning operation.

This implementation will run a `while` loop that reads the available bytes as they come in and spits them out to the `Sys.stdout`, and will continue until it gets an `EOF` error (git clone operation finished), which then it will read all of the bytes + exit code and return it as usual.


src/haxelib/api/Vcs.hx
```diff
@@ -193,6 +194,18 @@ abstract class Vcs implements IVcs {
 		// just in case process hangs waiting for stdin
 		p.stdin.close();
 
+		// In certain cases of git clones, it will hang on reading from stderr
+		// if we don't read from it. So we will always try to read from it.
```
tobil4sk, line 198: The reason this was solved with threads on windows is that the same thing could happen with p.stdout and p.stderr at the same time, in which case the only solution is to read from both simultaneously (with separate threads).

However, the threads caused issues on Linux so I made it windows only in: b8c24a783000c609ae3b45c27aab423d9e0d7193. I suspect the same issue is also behind these crashes: #591

tobil4sk: Neko has been patched so I have [reapplied](https://github.com/HaxeFoundation/haxelib/commit/649cd60b966b1b1e69cc8df2ae10ea2512298f76) that original fix for all platforms. That should handle the stderr/stdout issue.

PR #643, opened 2024-08-26
Allow compiling client with hxcpp target

tobil4sk:


client_cpp.hxml
```diff
@@ -0,0 +1,9 @@
+-cp src
+-cp hx3compat/std
+-cp hx4compat/std
+-cp crypto/src
```
Simn, line 4: Maybe we should take this opportunity to factor this out into a common hxml so that it's easier to change in the future.

PR #674, opened 2025-07-20
Implement automatic hxb caching for run scripts

tobil4sk:
This provides a nice improvement in responsiveness when running eval scripts, by making use of the hxb feature of haxe 5.

TODO:
- [ ] Flag to force a rebuild the cache
- [ ] Automatic handling of corrupt/outdated hxb cache

### DIFF
--- a/src/haxelib/api/ScriptRunner.hx
+++ b/src/haxelib/api/ScriptRunner.hx
@@ -47,9 +47,9 @@ typedef LibraryRunData = {
 }
 
 private enum RunType {
-	/** Runs compiled neko file at `path` **/
-	Neko(path:String);
-	/**  **/
+	/** Runs compiled neko file **/
+	Neko;
+	/** Compiles and runs script using haxe's eval target **/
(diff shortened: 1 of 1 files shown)


Simn:
Ah, nice! I stopped thinking about hxb scripts when it was brought up that this would come with distribution issues because hxb is tied to a compiler version, but supporting this kind of local pre-compilation is a neat idea.

tobil4sk:
> I stopped thinking about hxb scripts when it was brought up that this would come with distribution issues because hxb is tied to a compiler version

Yeah, being tied to a specific compiler version is a bit of an issue, especially when compared to neko which is pretty stable and is independent of haxe version. Having the option to run with a `run.hl` file might still be a good idea, especially for libraries that are intended to support older haxe versions.

If a library is already using an eval script, there is no reason not to apply this optimisation, although, we will have to come up with a proper mechanism to detect when the cache is mismatched with the current haxe version and requires recompilation. I was wondering if you had any ideas for this? How much of this should be handled in the compiler vs in haxelib?

PR #682, opened 2025-12-02
[SemVer] Reflect.compare() is not specified to return exactly 1 or -1…

kLabz:
… (or 0)

Still need to update tests

### DIFF
--- a/test/tests/TestSemVer.hx
+++ b/test/tests/TestSemVer.hx
@@ -62,6 +62,17 @@ class TestSemVer extends TestBase {
 		assertEquals( "invalid", parseInvalid("10.50.2-rc.01"));
 	}
 
+	public function testPreviewComparison() {
+		assertTrue(SemVer.ofString("0.1.1-beta.1") < SemVer.ofString("0.1.1"));
+		assertTrue(SemVer.ofString("0.1.1-beta.1") > SemVer.ofString("0.1.0"));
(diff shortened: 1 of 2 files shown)


test/tests/TestSemVer.hx
```diff
@@ -62,6 +62,17 @@ class TestSemVer extends TestBase {
...
 	}
 
+	public function testPreviewComparison() {
+		assertTrue(SemVer.ofString("0.1.1-beta.1") < SemVer.ofString("0.1.1"));
+		assertTrue(SemVer.ofString("0.1.1-beta.1") > SemVer.ofString("0.1.0"));
+		assertTrue(SemVer.ofString("0.1.1-alpha.2") > SemVer.ofString("0.1.1-alpha.1"));
+		assertTrue(SemVer.ofString("0.1.1-beta.1") > SemVer.ofString("0.1.1-alpha.2"));
+		assertTrue(SemVer.ofString("0.1.1-rc.1") > SemVer.ofString("0.1.1-beta.1"));
+		#if !(server || haxelib_api)
+		assertTrue(SemVer.ofString("0.1.1-preview.1") > SemVer.ofString("0.1.1-rc.1"));
```
tobil4sk, line 72: I don't think the bug was specific to preview versions, on hashlink this was evaluating to false:
```haxe
SemVer.ofString("1.2.2") < SemVer.ofString("1.2.4")
```
because the comparison returned `-2`

